PR #24, opened 2022-10-20
chore: remove unnecessary package-lock files

petebacondarwin:
Since we are using npm workspaces we only need  single package-lock at the root of the piercing directory.


dario-piotrowicz:
Ah yes, I noticed this after moving to workspaces but forgot to remove them, thanks! ❤️

PR #25, opened 2022-10-20
Automate-qwik-loader-import

petebacondarwin:
Previously we hard-coded this.


dario-piotrowicz:
mh... are you sure this is ok? when it is hardcoded we have a guarantee that the app won't suddenly break but now we wouldn't have that security and the app could break at any time (we would also get newer qwikLoaders while not updating our qwik versions increasingly the risk of the app stopping working even more).

if possible I would be much more comfortable if we could have an npm script which fetches the latest qwikLoader and saves it in a file (which you can then import in the same way you have it now) so that we have our own stable snapshots of the qwikLoader without the risk of getting the rug pulled from under our feet.

dario-piotrowicz:
@petebacondarwin sorry I misread the code! 😝 🙇

PR #29, opened 2022-10-20
refactor: clean up login fragment code

petebacondarwin:
This is a tentative PR just to see what it might look like.
I was playing around with trying to make the code look simpler.

@dario-piotrowicz - feel free to pick and choose what works for you here.

### DIFF
--- a/migration-strategies/piercing/app/fragments/login/src/root.tsx
+++ b/migration-strategies/piercing/app/fragments/login/src/root.tsx
@@ -3,123 +3,95 @@ import {
   component$,
   useRef,
   useStore,
-  useStyles$,
   useStylesScoped$,
 } from "@builder.io/qwik";
 import { dispatchPiercingEvent } from "piercing-library";
 
 import styles from "./root.css?inline";
 
+const enum InputErrorMessages {
+  noSpaces = "Spaces are not allowed",
+  noSpecialCharacters = "Special characters are not allowed",
+  maxLength = "The provided username is too long, it needs not to be longer than 20 characters",
+  emptyInput = "Please provide a username",
+}
+interface LoginState {
+  animatePassword: boolean;
+  usernameError: InputErrorMessages | null;
+  usernameTouched: boolean;
+  username: string;
+  password: string;
+  loading: boolean;
+}
+
+export const getUsernameInputError = (
+  input: string,
+  inputTouched: boolean
+): InputErrorMessages | null => {
+  if (input.includes(" ")) return InputErrorMessages.noSpaces;
+  if (/[^a-zA-Z0-9]/.test(input)) return InputErrorMessages.noSpecialCharacters;
+  if (input.length > 20) return InputErrorMessages.maxLength;
+  if (inputTouched && !input) return InputErrorMessages.emptyInput;
+  return null;
+};
+
 export const Root = component$(() => {
-  const possibleInputErrors = {
-    noSpaces: "noSpaces",
-    noSpecialCharacters: "noSpecialCharacters",
-    maxLength: "maxLength",
-    emptyInput: "emptyInput",
-  } as const;
-
-  const inputErrorMessages: Record<
-    typeof possibleInputErrors[keyof typeof possibleInputErrors],
-    string
-  > = {
-    [possibleInputErrors.noSpaces]: "Spaces are not allowed",
-    [possibleInputErrors.noSpecialCharacters]:
-      "Special characters are not allowed",
-    [possibleInputErrors.maxLength]:
-      "The provided username is too long, it needs not to be longer than 20 characters",
-    [possibleInputErrors.emptyInput]: "Please provide a username",
-  };
-
   const ref = useRef();
   useStylesScoped$(styles);
 
-  const inputsDetails = useStore<{
-    animatePasswordInput: boolean;
-    usernameInputError:
-      | null
-      | typeof possibleInputErrors[keyof typeof possibleInputErrors];
-    usernameInputTouched: boolean;
-  }>({
-    animatePasswordInput: false,
-    usernameInputError: null,
-    usernameInputTouched: false,
-  });
-
-  const userData = useStore({
+  const state = useStore<LoginState>({
+    animatePassword: false,
+    usernameError: null,
+    usernameTouched: false,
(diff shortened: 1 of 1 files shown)


migration-strategies/piercing/app/fragments/login/src/root.tsx
```diff
@@ -3,123 +3,95 @@ import {
...
-                value
-              );
-              setTimeout(
-                () => (inputsDetails.animatePasswordInput = false),
-                500
-              );
-            }}
-            value={userData.username}
+            onFocus$={handleFocus}
+            onInput$={handleInput}
```
petebacondarwin, line 89: I like the idea of moving handler code out into its own function because then that code can be considered separately. It can have more comments and so on without making the template look more messy. And if you really wanted you could write unit tests for the handlers.

migration-strategies/piercing/app/fragments/login/src/root.tsx
```diff
@@ -3,123 +3,95 @@ import {
...
-      | typeof possibleInputErrors[keyof typeof possibleInputErrors];
-    usernameInputTouched: boolean;
-  }>({
-    animatePasswordInput: false,
-    usernameInputError: null,
-    usernameInputTouched: false,
-  });
-
-  const userData = useStore({
+  const state = useStore<LoginState>({
```
petebacondarwin, line 42: In simple components like this I don't think there is much value in having multiple "stores".
If we wanted the code to be reusable, then I think we would move it into a custom hook instead.

migration-strategies/piercing/app/fragments/login/src/root.tsx
```diff
@@ -3,123 +3,95 @@ import {
...
 import styles from "./root.css?inline";
 
-export const Root = component$(() => {
-  const possibleInputErrors = {
-    noSpaces: "noSpaces",
-    noSpecialCharacters: "noSpecialCharacters",
-    maxLength: "maxLength",
-    emptyInput: "emptyInput",
-  } as const;
+const enum InputErrorMessages {
```
petebacondarwin, line 12: I felt a simple `const enum` was cleaner than doing that type dance with records to get the strings. WDYT?

migration-strategies/piercing/app/fragments/login/src/root.tsx
```diff
@@ -3,123 +3,95 @@ import {
...
+  noSpaces = "Spaces are not allowed",
+  noSpecialCharacters = "Special characters are not allowed",
+  maxLength = "The provided username is too long, it needs not to be longer than 20 characters",
+  emptyInput = "Please provide a username",
+}
+interface LoginState {
```
petebacondarwin, line 18: I think that free standing interfaces are quite nice since we can then add jsdoc comments to the members more easily than when it is inline.

dario-piotrowicz:
@petebacondarwin yesterday I rebased to the latest main... seems like that messed up your PR 😓, sorry about that 🙇 

I'm looking at your commit: https://github.com/cloudflare/workers-web-experiments/pull/29/commits/d5661dcbb5d8e51f2e3510e4e505da59e2a7b17d

we may want to rebase this PR or simply re-create a new one and cherry pick that commit? whatever works best for you 🙂

migration-strategies/piercing/app/fragments/login/src/root.tsx
dario-piotrowicz: yeah, I iterated through this code a couple of times myself 😅

Ultimately (if I remember correctly, since it was a while back 😅) I went with this solution because to me it provided the nicer type definitions, so that for example in the messages object you would see something like:
Screenshot 2022-10-21 at 11 24 11

Which to me looks clearer compared to what we would get with the `const enum`:
Screenshot 2022-10-21 at 11 23 45

(it would be great if the typescript language service could not convert the const enums to numbers at this stage 😅)

But I don't really have a strong strong opinion on this so if you prefer it I am totally fine with going with the `const enum` 🙂

dario-piotrowicz: ah... I was pretty sure `const enum` only worked with numbers and not strings 😮 

sure, your solution looks much better! 🙂

migration-strategies/piercing/app/fragments/login/src/root.tsx
dario-piotrowicz: yeah good point 🙂

migration-strategies/piercing/app/fragments/login/src/root.tsx
dario-piotrowicz: yeah.... I was just trying to have a logical distinction of the different states for the component (I wasn't doing it for performance or reusability reasons), it was sort of an attempt to make things more separated/clear, but I guess it may just raise more confusion than anything else 😖 

I'm totally ok with having only single "stores" 🙂

migration-strategies/piercing/app/fragments/login/src/root.tsx
dario-piotrowicz: mh.... yeah I do agree, my issue is that I feel this could generate more confusing code? like you've factored out `handleFocus` and `handleInput` but when you start reading the component's code you don't know who's input/focus you're handling (maybe it's the button's focus?).

so I feel like we should maybe clarify the names?

In general this is sort of my issue, if I want to separate a handler from the element itself, I feel like the name should be quite explanatory not to raise confusions, which, if there are various different elements/handlers this could get a bit out of hand.

Also if components are chunky this would make the code much much less clear I believe (but components should not get chunky in the first place I guess 😛).

Anyways sorry I started rambling a bit, I was just trying to say that I agree but in general I am a bit concerned as to if having the handlers on their own can make things less clear (in this case for example I would definitely prefer to rename the handlers, since the submit button can also get focus and a generic name here introduces ambiguity).

petebacondarwin: I agree with unambiguous naming being important.
But equally I find overly descriptive naming reduces readability of the code.
So where possible I try to keep names as simple as possible.
In this case the whole template is visible on a single screen (for me at least) so it is easy to see what these handlers are handling.

PR #30, opened 2022-10-21
Show seams in piercing demo

jculvey:


jculvey:
image

cloud-gallery/header/src/Slider/Slider.tsx
```diff
@@ -1,6 +1,13 @@
...
+  component$,
+  useStore,
+  useStylesScoped$,
+  $,
+  useClientEffect$,
+  useRef,
+} from "@builder.io/qwik";
 import { getCookie, saveCookie } from "helpers";
-import styles from "./Slider.css?inline";
+import styles from "./Slider.scss?inline";
```
dario-piotrowicz, line 10: scss??? is that supported by qwik? 😮

cloud-gallery/header/src/Slider/Slider.scss
```diff
@@ -26,3 +27,13 @@
...
+
+.seam-options-container {
+	.label {
+		margin-left: .5rem;
+	}
+
+	input {
+		margin-right: .5rem;
+	}
+}
```
dario-piotrowicz, line 39:
```suggestion
}

```

dario-piotrowicz:
Could the seams be of different colors depending on the fragment? 🙂

Also in the piercing demo the root one shouldn't say "fragment" but something along the lines or "legacy app" since that is not actually a fragment

Also I have a feeling that the text is slightly too big, I am wondering if we can make it a bit smaller and unobtrusive (maybe without a background? 🤔)

(PS: if you want to move to something else I can make the changes in the piercing demo 🙂)

cloud-gallery/header/src/Slider/Slider.tsx
jculvey: Indeed it is :)

migration-strategies/piercing/app/productivity-suite/index.html
```diff
@@ -96,6 +96,34 @@
         border-left: 1px solid currentColor;
         padding-left: 0.5rem;
       }
+
+      body.show-seams #root,
```
petebacondarwin, line 100: Ah! That's how you do it! Nice TIL.

petebacondarwin (changes requested):
We should not delete `package-lock.json` it is essential for ensuring that we can get the same result for a particular commit for ever.

cloud-gallery/header/src/Slider/Slider.tsx
dario-piotrowicz: 😮 👍

jculvey:
Ok I added back the package-lock.json and tweaked the seam styles. Here's how it's looking now:

Screen Shot 2022-10-21 at 2 04 05 PM

Screen Shot 2022-10-21 at 2 01 56 PM

dario-piotrowicz:
Awesome! thanks @jculvey 😍

PR #37, opened 2022-10-26
Message Bus Implementation

dario-piotrowicz:


productivity-suite/app/fragments/login/src/components/LoginForm.tsx
```diff
@@ -49,13 +49,16 @@ export const LoginForm = component$(() => {
...
-      payload: {
+    getBus(ref.value!).dispatch(
+      "login",
+      {
         username: state.username,
         password: state.password,
       },
-    });
+      {
+        singleUse: true,
```
dario-piotrowicz, line 59: I had to add this `singleUse` to make sure that some events get consumed only once

For example the `login` event, if we keep the last login event in the bus, then we wouldn't be able to logout, since on logout the app would go back to the login page, listen for the `login` event, find it in the bus and automatically re-log in.

productivity-suite/app/fragments/todo-lists/src/dispatchSelectedListUpdated.ts
```diff
@@ -1,16 +1,11 @@
-import { dispatchPiercingEvent } from "piercing-library";
+import { getBus } from "piercing-library";
 import type { TodoList } from "shared";
 
 export function dispatchSelectedListUpdated(
   el: Element,
-  listSelected: TodoList,
-  which?: "previous" | "next"
```
dario-piotrowicz, line 7: I've totally remove the animation logic for now, I am considering if we do want it or not (since now the todos fragment doesn't get removed and added, so it looks nice even without any entering/leaving animation)

productivity-suite/app/fragments/todos/src/App.tsx
```diff
@@ -32,6 +38,16 @@ const App: React.FC<{
 
   if (!currentUser || !listName) return <></>;
 
+  function setTodos(todos: Todo[], dispatch = true) {
+    setTodosState(todos);
```
dario-piotrowicz, line 42: `setTodos` and `setTodosState`.... we may need better names 😓

productivity-suite/piercing-library/src/message-bus.ts
```diff
@@ -0,0 +1,113 @@
...
+      }
+    }
+    this.handlers.push(handler);
+    return () => {
+      this.handlers = this.handlers.filter((h) => h !== handler);
+    };
+  }
+}
+
+// export class ServerSideMessageBus implements MessageBus {
```
dario-piotrowicz, line 49: all this commented out code is to be introduced for the Server side part of the message bus

productivity-suite/piercing-library/src/piercing-fragment-host/piercing-fragment-host.ts
```diff
@@ -1,108 +1,100 @@
...
 // Important: we only need to import the type, do not import the class itself
 //            as it would bloat the script (which needs to be very light)
 import type { PiercingFragmentOutlet } from "../piercing-fragment-outlet";
 
-export class PiercingFragmentHost extends HTMLElement {
-	fragmentId!: string;
-	piercingEventsQueue: CustomEvent<PiercingEvent>[] = [];
-	queueEventListener?: (event: Event) => void;
-	stylesEmbeddingObserver?: MutationObserver;
+export class PiercingFragmentHost extends HTMLElement implements MessageBus {
```
dario-piotrowicz, line 12: @petebacondarwin instead of having a `FragmentMessageBus` field here I just figured the `PiercingFragmentHost` can implement the `MessageBus` interface itself, do you see some possible issues with it?

productivity-suite/piercing-library/src/piercing-fragment-host/piercing-fragment-host.ts
```diff
@@ -1,108 +1,100 @@
...
-	constructor() {
-		super();
-	}
+  constructor() {
+    super();
+  }
 
-	async connectedCallback() {
-		this.fragmentId = this.getAttribute("fragment-id")!;
+  private messageHandlerRemovers: (() => void)[] = [];
```
dario-piotrowicz, line 22: @petebacondarwin I kept the `Remover` name, I wonder if we can find a more descriptive name? maybe `messageHandlerCleanupFunctions` or something like that? 🤔

productivity-suite/app/fragments/login/src/components/LoginForm.tsx
petebacondarwin: What if we had an `authentication` event, which we use for both login and logout? Then we could send:

```ts
// Login
...dispatch("authentication", {username, password})
```

```ts
// Logout
...dispatch("authentication", null)
```

productivity-suite/app/fragments/todo-lists/src/root.tsx
```diff
@@ -44,17 +46,36 @@ export const Root = component$(() => {
...
+          listName: string;
+          todos: Todo[];
+        }) => {
+          const list = state.todoLists.find(({ name }) => name === listName);
+          if (list) {
+            list.todos = todos;
+          }
+        },
+      });
+    }
```
petebacondarwin, line 66: Why is this listener here?

productivity-suite/app/fragments/login/src/components/LoginForm.tsx
dario-piotrowicz: ah, yeah that didn't cross my mind 😅

sounds like a nice solution, I'll apply that change 🙂👍

productivity-suite/piercing-library/src/piercing-fragment-host/piercing-fragment-host.ts
petebacondarwin: How does `getBus()` find this since it is looking for elements that have the `messageBus` property? I think we are safer keeping it separate.
In the future, we could also create other kinds of MessageBus instances if we needed (not that I can think of a use-case right now)

productivity-suite/app/fragments/todo-lists/src/root.tsx
dario-piotrowicz: ah yeah, good spot! 😁 

so basically we want the `todo-lists` to be up to date with the todos, because when you select a list, it dispatches an event with the whole list and its todos, which is then taken by the todos fragment which uses that data. If the `todo-lists` wouldn't keep the todos up to date then the todos fragment would receive old outdated todos when lists are selected.

I found this to be a sort of ok, I also thought of other solutions (which seemed slightly worse to me):
- having the `todo-list-selected` only include the list name and then whoever needs the todos would manually "fetch them from the backend" (/take them from the cookies)
- having the todos fragment keep the whole list of lists in its state (currently it only keeps the current one since it has no need for the rest), so that it can ignore the todos coming from the even and use those that is has in its state (the event can also only include the list name here as well)

what do you think? does one of the other solution look better to you? or do you have a different idea/suggestion? 🙂

productivity-suite/piercing-library/src/message-bus.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (tmpEl?.messageBus) return tmpEl.messageBus;
+
+  const gThis = globalThis as WithOptionalMessageBus<typeof globalThis>;
+  if (!gThis.messageBus) {
+    // TODO: on the server this needs to be the server side message bus!
+    gThis.messageBus = new BrowserMessageBus();
+  }
+
+  return gThis.messageBus;
+}
```
petebacondarwin, line 76: How about something like this?

```ts
export function getBus(element: Element): MessageBus {
  let tmpEl: Element | null = element;
  while (tmpEl) {
    if (hasMessageBus(tmpEl)) {
      return tmpEl.messageBus;
    }
    tmpEl = tmpEl.parentElement;
  }
  return ensureMessageBus(globalThis).messageBus;
}

function hasMessageBus(obj: unknown): obj is { messageBus: MessageBus } {
  return !!obj && "messageBus" in obj;
}

function ensureMessageBus<T extends object>(
  obj: T
): T & { messageBus: MessageBus } {
  if (hasMessageBus(obj)) {
    return obj;
  } else {
    const objWithBus = obj as T & { messageBus: MessageBus };
    objWithBus.messageBus = new BrowserMessageBus();
    return objWithBus;
  }
}
```

petebacondarwin: Actually let's go a step further and throw an error if globalThis does not contain a MessageBus.

productivity-suite/app/fragments/todo-lists/src/root.tsx
petebacondarwin: It feels like the `todos-list` fragment should only care about the todo list names.
And the `todos` fragment should only care about the todo items in the list it is displaying.

productivity-suite/app/fragments/todos/src/App.tsx
petebacondarwin: Does React State work a bit different to Qwik Store?
If we bundle these together will it trigger more rendering changes that we want?

productivity-suite/piercing-library/src/message-bus.ts
petebacondarwin:
```ts
export function getBus(element: Element): MessageBus {
  let tmpEl: Element | null = element;
  while (tmpEl) {
    if (hasMessageBus(tmpEl)) {
      return tmpEl.messageBus;
    }
    tmpEl = tmpEl.parentElement;
  }
  if (hasMessageBus(globalThis)) {
    return globalThis.messageBus;
  }
  throw new Error("No global message bus defined!");
}

function hasMessageBus(obj: unknown): obj is { messageBus: MessageBus } {
  return !!obj && "messageBus" in obj;
}
```

dario-piotrowicz: I am not sure I follow, if `globalThis` doesn't have the message bus shouldn't we just create it? (isn't that how it gets created in the first place?)

productivity-suite/app/fragments/todos/src/App.tsx
dario-piotrowicz: no, the issue is that actually `setTodos` simply calls `setTodosState` but also dispatches the event, actually I think we can just keep the original `setTodos` (instead of renaming it to `setTodosState`) and call the other function `setTodosAndDispatchUpdate` or something along those lines?

productivity-suite/app/fragments/todo-lists/src/root.tsx
petebacondarwin: Yeah, I think so. Micro-frontends should be pretty self-contained and not really rely on other fragments (or even their container) to provide them with data.

petebacondarwin: There could be some cleverness done with Service Workers / Durable Objects / Caching / GraphQL that might avoid duplicated fetches...

productivity-suite/piercing-library/src/message-bus.ts
petebacondarwin: We will need to initialize the message bus with events from SSR, so I think it is cleaner to create it in a script that is injected into the SSR response.

productivity-suite/app/fragments/todos/src/App.tsx
dario-piotrowicz: I've refactored this bit so now there's only `setTodos` 🙂👍

productivity-suite/app/fragments/todo-lists/src/root.tsx
dario-piotrowicz: I've moved changed the events so that only the list name is being dispatched and the todos are fetched if needed 🙂, so it should be all good 🙂

productivity-suite/piercing-library/src/message-bus.ts
dario-piotrowicz: I've applied your latest suggestion 🙂👍

productivity-suite/piercing-library/src/piercing-fragment-host/piercing-fragment-host.ts
dario-piotrowicz: fixed! 🙂 

the cleaning up does seem to work nicely too! 🤩

Screenshot 2022-10-31 at 22 23 14

productivity-suite/piercing-library/src/message-bus/message-bus.ts
```diff
@@ -0,0 +1,54 @@
...
+  listen(handler: MessageHandler): (() => void) | null;
+}
+
+export class GenericMessageBus implements MessageBus {
+  protected _handlers: MessageHandler[] = [];
+
+  constructor(protected _state: MessageBusState = {}) {}
+
+  get state(): MessageBusState {
+    return JSON.parse(JSON.stringify(this._state));
```
dario-piotrowicz, line 21: overkill? 🤔

productivity-suite/piercing-library/src/message-bus/inline-script.ts
```diff
@@ -0,0 +1,4 @@
+import { createBrowserMessageBusFromStateStr } from "./browser-message-bus";
+
+const stateStr = "__MESSAGE_BUS_STATE_STR_PLACEHOLDER__";
```
dario-piotrowicz, line 3: using a placeholder and replacing it when we apply the script was the only way I could think of to have the inline script built and apply the state when we add actually make use of the script.

alternatively I think we could also have the `stateStr` read from some global variable and update that variable before using the script...

does anyone has a better solution? 🙂

productivity-suite/app/fragments/todos/src/main.tsx
```diff
@@ -1,26 +1,25 @@
...
-  const match = /\/todos\/([^/]+)/.exec(window.location.pathname);
-  const listName = match?.[1] ?? null;
-  let todoList = null;
+  const rootElement = document.getElementById(
+    "todos-fragment-root"
+  ) as HTMLElement;
+
+  const { name: todoListName }: { name: string } =
+    getBus().latestValue("todo-list-selected");
+  const { username: currentUser }: { username: string } =
```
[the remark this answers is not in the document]
dario-piotrowicz: good point! 👍

productivity-suite/app/fragments/todos/src/main.tsx
```diff
@@ -31,10 +30,6 @@ import { EnvContext } from "./env";
     </React.StrictMode>
   );
 
-  const rootElement = document.getElementById(
-    "todos-fragment-root"
-  ) as HTMLElement;
```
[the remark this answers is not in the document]
dario-piotrowicz: what do you mean? 😁 

By the way, I've just moved it above so that I can use it in the `getBus` (which I am not actually doing 😅)

productivity-suite/app/fragments/todos/src/main.tsx
```diff
@@ -1,26 +1,25 @@
...
 (async () => {
-  const match = /\/todos\/([^/]+)/.exec(window.location.pathname);
-  const listName = match?.[1] ?? null;
-  let todoList = null;
+  const rootElement = document.getElementById(
+    "todos-fragment-root"
+  ) as HTMLElement;
+
+  const { name: todoListName }: { name: string } =
+    getBus().latestValue("todo-list-selected");
```
dario-piotrowicz, line 14:
```suggestion
    getBus(rootElement).latestValue("todo-list-selected");
```

productivity-suite/app/legacy-app/src/components/pages/TodoLists.tsx
```diff
@@ -1,67 +1,61 @@
...
+            if (name !== selectedListName) {
+              const previousNameNotProvided = !selectedListName;
+              setSelectedListName(name);
+              setTimeout(() => {
+                if (!noNavigation) {
+                  navigate(`/todos/${name}`, {
+                    replace: previousNameNotProvided,
+                  });
+                }
+              }, 1);
```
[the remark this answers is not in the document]
dario-piotrowicz: yes, this is the same exact thing we had with the customEvents, if the legacy app renders the component and it triggers the navigation right away react complains.

ok, then should I add it to the message bus implementation instead? I guess it should be configurable like we did for the custom events?

productivity-suite/piercing-library/src/message-bus/fragment-message-bus.ts
```diff
@@ -0,0 +1,38 @@
...
+  dispatch(eventName: string, value: any) {
+    return this.parentBus.dispatch(eventName, value);
+  }
+
+  listen(eventName: string, callback: MessageBusCallback) {
+    const callbackRemover = this.parentBus.listen(eventName, callback);
+    if (callbackRemover) {
+      this.callbackRemovers.push(callbackRemover);
+    }
+    return callbackRemover;
```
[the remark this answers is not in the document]
dario-piotrowicz: ahhhh good point, I am creating a new/different memory leak here 😓👍

productivity-suite/piercing-library/src/message-bus/message-bus.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export type MessageBusState = Record<string, any>;
+
+export interface MessageBus {
+  state: MessageBusState;
+  dispatch(eventName: string, value: any): void;
+  listen(
+    eventName: string,
+    callback: (value: any) => void
+  ): (() => void) | null;
```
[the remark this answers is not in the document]
dario-piotrowicz: ah yeah! 😓👍

productivity-suite/piercing-library/src/message-bus/message-bus.ts
```diff
@@ -0,0 +1,51 @@
...
+  constructor(protected _state: MessageBusState = {}) {}
+
+  get state(): MessageBusState {
+    return JSON.parse(JSON.stringify(this._state));
+  }
+
+  dispatch(eventName: string, value: any) {
+    this._state[eventName] = value;
+    const callbacksForEvent = this._callbacksMap.get(eventName) ?? [];
+    callbacksForEvent.forEach((callback) => callback(value));
```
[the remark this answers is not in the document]
dario-piotrowicz: good point! 👍

productivity-suite/piercing-library/src/piercing-gateway.ts
```diff
@@ -211,11 +249,9 @@ export class PiercingGateway<Env> {
     streamToInclude: ReadableStream
   ): Promise<Response> {
     const baseUrl = this.config.getBaseAppUrl(env).replace(/\/$/, "");
-    const newRequest = new Request(baseUrl, {
-      ...request,
-      headers: { Accept: "text/html" },
-    });
-    const indexHtmlResponse = await this.fetchBaseIndexHtml(newRequest);
+    const updatedRequest = new Request(baseUrl, request);
+    updatedRequest.headers.set("accept", "text/html");
```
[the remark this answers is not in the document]
dario-piotrowicz: yeah.... to be honest I can't totally remember why I added it.... maybe as an extra (and unnecessary) precaution..? 🤷 

I'll remove it 🙂👍

productivity-suite/piercing-library/src/piercing-fragment-host/piercing-fragment-host.ts
```diff
@@ -1,108 +1,82 @@
...
+import { messageBusProp } from "../message-bus/message-bus-prop";
+import { FragmentMessageBus } from "../message-bus/fragment-message-bus";
 
-	async connectedCallback() {
-		this.fragmentId = this.getAttribute("fragment-id")!;
+export class PiercingFragmentHost extends HTMLElement {
+  [messageBusProp]: FragmentMessageBus;
 
-		this.setQueueEventListener();
+  fragmentId!: string;
```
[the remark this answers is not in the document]
dario-piotrowicz: feels a bit hacky 😅

also if the fetch fails we get this error message:
Screenshot 2022-11-01 at 15 13 20

having that Id would produce something like:
`The fragment with id "ERROR MISSING FRAGMENT ID!" is not present and it could not be fetched`
instead of:
`The fragment with id "undefined" is not present and it could not be fetched`

WDYT?

productivity-suite/piercing-library/src/message-bus/message-bus.ts
```diff
@@ -0,0 +1,51 @@
...
+    const latestValue = this.latestValue(eventName);
+    if (latestValue) {
+      callback(latestValue);
+    }
+    if (!this._callbacksMap.has(eventName)) {
+      this._callbacksMap.set(eventName, []);
+    }
+    this._callbacksMap.get(eventName)!.push(callback);
+    return () => {
+      const callbacks = this._callbacksMap.get(eventName)!;
```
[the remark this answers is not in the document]
dario-piotrowicz: I see, I didn't thought this were too important since a key in the map is quite minor and also given the removers' implementation calling one multiple times results in practically a noop (but only because of the current implementation!)

I'll fix it thanks 🙂

productivity-suite/app/fragments/todos/package.json
```diff
@@ -6,7 +6,7 @@
 	"type": "module",
 	"scripts": {
 		"dev:vite": "vite dev -c vite.config.client.ts",
-		"build:client": "tsc && vite build -c vite.config.client.ts",
+		"build:client": "tsc && vite build -c vite.config.client.ts -m development",
```
dario-piotrowicz, line 9: oops, added by mistake 😅

productivity-suite/piercing-library/src/message-bus/fragment-message-bus.ts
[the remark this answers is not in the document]
dario-piotrowicz: ah yeah, no, there is not a complete memory leak here since we have the `clearAllHandlers` below, but still things should be cleaned up 🙂

productivity-suite/piercing-library/src/piercing-fragment-host/piercing-fragment-host.ts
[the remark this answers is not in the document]
dario-piotrowicz: as we discussed there's already a check for the id:
Screenshot 2022-11-02 at 10 50 12

(and I've added the same check in the host component which was missing it)

so that should be ok?

productivity-suite/piercing-library/src/message-bus/message-bus.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  dispatch(eventName: string, value: any) {
+    this._state[eventName] = value;
+    const callbacksForEvent = this._callbacksMap.get(eventName) ?? [];
+    callbacksForEvent.forEach((callback) => callback(value));
+  }
+
+  listen(eventName: string, callback: MessageBusCallback) {
+    const latestValue = this.latestValue(eventName);
+    if (latestValue) {
```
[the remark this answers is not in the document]
dario-piotrowicz: With the addition of the `MessageBusValue` I changed this check to be `latestValue !== undefined` since undefined is not a valid `MessageBusValue`:
Screenshot 2022-11-02 at 11 44 04

I hope that's ok? 🙂

productivity-suite/piercing-library/src/message-bus/inline-script.ts
```diff
@@ -0,0 +1,4 @@
...
+createBrowserMessageBusFromStateStr(stateStr);
```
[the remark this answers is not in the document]
dario-piotrowicz: I've applied the change anyways since I do like it 😄

productivity-suite/piercing-library/src/message-bus/message-bus.ts
dario-piotrowicz: @petebacondarwin with your new types you've suggested the fact that the state is serializable is guaranteed by typescript (unless someone hacks around).

So do you still agree to keep this in? 🙂

productivity-suite/piercing-library/src/message-bus/inline-script.ts
dario-piotrowicz: Better solution suggested by Pete 🙂

productivity-suite/piercing-library/src/message-bus/message-bus.ts
petebacondarwin: Yeah let's keep this in for now.

productivity-suite/app/fragments/login/src/components/LoginForm.tsx
```diff
@@ -49,12 +49,9 @@ export const LoginForm = component$(() => {
   });
 
   const dispatchLoginEvent = $(() => {
-    dispatchPiercingEvent(ref.value!, {
-      type: "login",
-      payload: {
-        username: state.username,
-        password: state.password,
-      },
+    getBus(ref.value!).dispatch("authentication", {
```
petebacondarwin, line 52: The ! is not actually needed from a type checking point of view,
since getBus() can accept undefined.

Although I think that would be a bug. So the safest thing would be to check that `ref.value` is not undefined and throw an error if it is, rather than just telling TS to ignore the undefined typing.

productivity-suite/app/fragments/todo-lists/src/dispatchSelectedListUpdated.ts
```diff
@@ -1,16 +1,7 @@
...
-) {
-  dispatchPiercingEvent(el, {
-    type: "todo-list-selected",
-    payload: {
-      list: listSelected,
-      which,
-    },
+export function dispatchSelectedListUpdated(el: Element, listName: string) {
```
petebacondarwin, line 3: NIT, if you name the param `name` then you can write:

```ts
getBus(el).dispatch("todo-list-selected", {name});
```

productivity-suite/app/fragments/todo-lists/src/root.tsx
```diff
@@ -31,30 +28,76 @@ export const Root = component$(() => {
...
+            if (newIdxOfSelectedList >= 0) {
+              state.idxOfSelectedList = newIdxOfSelectedList;
+              state.selectedListName =
+                state.todoLists[newIdxOfSelectedList].name;
+            }
+          }
+        );
+      }
+    },
+    { eagerness: "load" }
```
petebacondarwin, line 77: Maybe add a comment here explaining this workaround?

dario-piotrowicz: yeah I thought about that too at some point (but I wasn't sure if it would add noise) 😕

I'll add it 🙂👍

productivity-suite/app/fragments/todo-lists/src/dispatchSelectedListUpdated.ts
dario-piotrowicz: yeah...

the reason why I did it in this way is that in `dispatchSelectedListUpdated` I feel like `name` is slightly too generic but I don't want to use `listName` in the `todo-list-selected` payload as there on the other hand it would be too specific.

I think I kinda prefer it this way, but if you do quite prefer the `name` I'm ok with it 🙂

dario-piotrowicz: PS: what do you think if I just move this function in the root of the todo-lists component (which is the only place that uses it)?

productivity-suite/app/fragments/login/src/components/LoginForm.tsx
dario-piotrowicz: good point, I'm pretty sure I added the `!` when we were accepting `Element | null` so `undefined` was actually causing typing issues, and missed it during the refactoring 😅 

I agree that now we can just do `getBus(ref.value)`

I don't think that it can ever be `undefined` since we're just applying it below in the jsx (and this is not a generic/reusable code), do you totally think that a check for that is needed?

productivity-suite/app/fragments/todo-lists/src/root.tsx
dario-piotrowicz: I've added a comment here and in the outlet component when we emit the `qinit`, I kept the comments short and added a link to the issue you've opened, please let me know if you think it's ok 🙂

productivity-suite/app/fragments/todo-lists/src/root.tsx
```diff
@@ -31,30 +28,76 @@ export const Root = component$(() => {
...
+    const selectedListName =
+      getBus().latestValue<{ name: string }>("todo-list-selected")?.name ??
+      null;
+
+    const idx = state.todoLists.findIndex(
+      ({ name }) => name === selectedListName
+    );
+
+    state.idxOfSelectedList = idx !== -1 ? idx : state.todoLists.length - 1;
+    state.selectedListName = state.todoLists[idx]!.name;
```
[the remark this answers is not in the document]
dario-piotrowicz: done 🙂

productivity-suite/app/fragments/todo-lists/src/dispatchSelectedListUpdated.ts
dario-piotrowicz: done both 👍

productivity-suite/app/fragments/login/src/components/LoginForm.tsx
dario-piotrowicz: I've removed the `!`, please let me know if you think I also need to add the check 🙂

productivity-suite/app/legacy-app/src/components/pages/TodoLists.tsx
```diff
@@ -1,67 +1,59 @@
...
-    "previous" | "next" | undefined
-  >();
-  const [showTodos, setShowTodos] = useState<boolean>(true);
+  const location = useLocation();
 
   useEffect(() => {
-    const match = /\/todos\/([^/]+)/.exec(window.location.pathname);
-    if (match) {
-      setSelectedListName(decodeURIComponent(match[1]));
+    const pathname = location.pathname;
```
dario-piotrowicz, line 13: @petebacondarwin I put back this change, what was I thinking 🤦 , you definitely (unfortunately) need to get the list name from the location because the one in the bus will not get updated 😕 
(I made that change earlier today so I think I flat out broke it but didn't notice it 😓 )

jculvey (approved):

productivity-suite/piercing-library/src/message-bus/message-bus-prop.ts
```diff
@@ -0,0 +1 @@
+export const messageBusProp = Symbol.for("fragment-message-bus");
```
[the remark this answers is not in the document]
dario-piotrowicz: no, it's because of the following code:
Screenshot 2022-11-04 at 15 46 17

That code is in the host web component which gets converted into an inline script, so in order not to slow down thing we need to try to make the inline script as small as possible, if i had:
`import { messageBusProp } from '../message-bus/message-bus.ts'` all the content of the message-bus file would also be added to the inline script (which is not that much but I was trying to keep the inline script as small as possible). So by splitting things in multiple files we can get manage to get into the inline script only what's really really necessary.

I hope it makes sense, let me know if you see some issues with that 🙂

PR #38, opened 2022-10-27
fix(productivity-suite): ensure the todos fragment has an id for hydration

petebacondarwin:
A previous commit (of mine) broke the rehydration of this fragment by stripping
the div that was used as the marker for the root of the component to rehydrate.

This fix moves the id onto the root of the `App` component itself.

### DIFF
--- a/productivity-suite/app/fragments/todos/src/worker/index.tsx
+++ b/productivity-suite/app/fragments/todos/src/worker/index.tsx
@@ -46,9 +46,11 @@ export default {
       env
     ).then((res) => res.text());
 
-    const [pre, post] = initialPageHtml.split(
-      '<div id="todos-fragment-root"></div>'
-    );
+    const appRootElement = '<div id="todos-fragment-root">';
+    const indexOfFragmentDiv = initialPageHtml.indexOf(appRootElement);
+    const indexForFragment = indexOfFragmentDiv + appRootElement.length;
+    const pre = initialPageHtml.slice(0, indexForFragment);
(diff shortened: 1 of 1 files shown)


productivity-suite/app/fragments/todos/src/App.tsx
```diff
@@ -63,7 +63,7 @@ const App: React.FC<{
   }
 
   return (
-    <div className="todo-mvc-wrapper">
+    <div id="todos-fragment-root" className="todo-mvc-wrapper">
```
dario-piotrowicz, line 66: is this the right fix? 😮 

this is the bit that uses this id:
Screenshot 2022-10-27 at 11 32 07

so basically you're creating the `App` and inside it you create this div with the `todos-fragment-root` id and then you look for this div in order to know where to create/hydrate the application (which is the one creating the div in the first place)?

am I missing something? 🤔

dario-piotrowicz: also locally the app does seem to be complaining about it?
Screenshot 2022-10-27 at 11 37 15

petebacondarwin: Doh! I really am not doing well here, eh?
I'll just unravel most of these changes...

dario-piotrowicz: ah no worries, it happens, I guess you could have missed the console log and react just shows the thing nicely anyways 😅

dario-piotrowicz (approved):

PR #39, opened 2022-10-27
feat(piercing) Add news tab using solid hackernews demo

jculvey:
image


productivity-suite/app/fragments/news/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "solid-hackernews2",
+  "scripts": {
+    "dev": "solid-start dev",
+    "build": "solid-start build",
+    "start": "npx wrangler dev --local --env local"
```
dario-piotrowicz, line 6: I don't see a local environment in the toml file so I guess this `--env local` is not actually needed?

productivity-suite/app/fragments/news/vite.config.ts
```diff
@@ -0,0 +1,42 @@
...
+  return {
+    name: "add-default-fn-export-to-bundle-plugin",
+    enforce: "post",
+    generateBundle(_options, chunkMap) {
+      for (const file of Object.keys(chunkMap)) {
+        if (/entry-client\..*\.js$/.test(file)) {
+          const jsChunk = chunkMap[file] as { code: string; isEntry: boolean };
+          if (jsChunk.isEntry) {
+            jsChunk.code = `
+							const moduleFn = () => {
```
dario-piotrowicz, line 15: the formatting is all wrong.... I'm pretty sure it is my fault 😓 , could you fix it please? (also in the other vite config 🙏 )

productivity-suite/app/fragments/news/vite.config.ts
```diff
@@ -0,0 +1,42 @@
+import solid from "solid-start/vite";
+import cloudflareWorkers from "solid-start-cloudflare-workers";
+import { defineConfig, PluginOption } from "vite";
+
+function addDefaultFnExportToBundle(): PluginOption {
```
dario-piotrowicz, line 5: this is just a thought, but could we move this function in the `shared` library so that we don't need to re-define it each time? (it can take the regex as a parameter).

productivity-suite/app/fragments/news/vite.config.ts
dario-piotrowicz: (please make sure that the `replace` at the end keeps working nicely after the formatting fix 🙂)

productivity-suite/app/legacy-app/src/components/layout.css
```diff
@@ -1,206 +1,221 @@
 .app-header {
-    display: flex;
```
dario-piotrowicz, line 2: are most of the changes in this file caused by auto-formatting? We haven't actually configured prettier in this demo/repo, I guess we should? what do you think @petebacondarwin?

productivity-suite/app/legacy-app/src/worker/script.ts
```diff
@@ -173,4 +173,41 @@ gateway.registerFragment({
...
+gateway.registerFragment({
+  fragmentId: "news",
+  // Note: deployment part of the url is fine also for local development since then
+  //       only the path part of the url is being used
+  getBaseUrl: () =>
+    `https://productivity-suite-news-fragment.devdash.workers.dev`,
+  prePiercingStyles: `
+    :not(piercing-fragment-outlet) > piercing-fragment-host[fragment-id="news"] {
+      position: absolute;
+      top: 24.6rem;
```
dario-piotrowicz, line 185: I'm sure this needs to be tweaked for various screensizes, it is a bit of a fidgety process, let me know if you'd like me to take care of it 🙂

productivity-suite/app/legacy-app/src/worker/script.ts
```diff
@@ -173,4 +173,41 @@ gateway.registerFragment({
...
+  shouldBeIncluded: async (request: Request) =>
+    isPiercingEnabled(request) &&
+    (await isUserAuthenticated(request)) &&
+    /^\/(news(\/[^/]+)?)?$/.test(new URL(request.url).pathname),
+  transformRequest: (
+    request: Request,
+    env: Env,
+    thisConfig: FragmentConfig<Env>
+  ) => {
+    console.log("news from the gateway");
```
dario-piotrowicz, line 199:
```suggestion
```

productivity-suite/app/fragments/news/src/types.ts
```diff
@@ -0,0 +1,19 @@
+export interface IComment {
```
dario-piotrowicz, line 1: I don't think that in typescript there is a convention to add an `I` in front of interface names, it feels strange to me and I haven't done this in the rest of the interfaces in the demo, what do we think? is there a real value in adding the `I` prefix?

productivity-suite/app/fragments/news/src/root.tsx
```diff
@@ -0,0 +1,58 @@
+// @refresh reload
```
dario-piotrowicz, line 1: out of curiosity, what does this do? 😮

dario-piotrowicz:
Besides my other comments I think the styles could/should be improved, I've seen lots of `px` being used which would make this fragment the only non-font-responsive part of the application.

Please let me know if you'd like me to improve this aspect as I'd gladly take a crack at it.

productivity-suite/app/legacy-app/src/components/layout.css
jculvey: yeah, they are. the significant changes were to the footer to make it sticky.

yes +1 to prettier or disregard formatting completely in review.

productivity-suite/app/legacy-app/src/worker/script.ts
jculvey: I didn't touch the CSS at all. I was trying to keep the example as is as much as possible. If you want to tweak it go to town.

productivity-suite/app/fragments/news/src/types.ts
jculvey: Copy pasted from the example repo

productivity-suite/app/fragments/news/src/root.tsx
jculvey: copy pasted from the example repo, no clue

jculvey:
> Besides my other comments I think the styles could/should be improved, I've seen lots of `px` being used which would make this fragment the only non-font-responsive part of the application.
> 
> Please let me know if you'd like me to improve this aspect as I'd gladly take a crack at it.

I didn't touch the CSS at all. I was trying to keep the example as is as much as possible. If you want to tweak it go to town.

productivity-suite/app/fragments/news/src/root.tsx
dario-piotrowicz: in these cases I think it would be better to get rid of it, if it breaks something we'll (hopefully) notice, much better than having there a piece of code that we don't know what it does 😅

productivity-suite/app/fragments/news/src/types.ts
dario-piotrowicz: I see, I am not a big fan but if we want to keep it as similar to the actual solid app as possible we could keep it like this...

I'd be curious to see if @petebacondarwin has any view on it 🙂

productivity-suite/app/fragments/news/package.json
```diff
@@ -0,0 +1,28 @@
...
+  },
+  "type": "module",
+  "main": "./dist/index.js",
+  "devDependencies": {
+    "solid-start-cloudflare-workers": "^0.2.0",
+    "solid-start-node": "^0.2.0",
+    "typescript": "^4.8.4",
+    "vite": "^3.1.8",
+    "wrangler": "^2.1.13",
+		"shared": "file:../../shared"
```
dario-piotrowicz, line 16:
```suggestion
   "shared": "file:../../shared"
```

dario-piotrowicz, line 16: this is misaligned 🙈

productivity-suite/app/fragments/news/src/root.tsx
```diff
@@ -0,0 +1,58 @@
...
+        <Suspense fallback={<div class="news-list-nav">Loading...</div>}>
+          <Routes>
+            <Route path="/*" component={News} />
+          </Routes>
+        </Suspense>
+      </ErrorBoundary>
+
+      {isSSR && <HydrationScript />}
+
+      {isServer && (
```
dario-piotrowicz, line 49: what's the different between `isSSR` and `isServer`? 😮

productivity-suite/app/fragments/news/src/root.css
```diff
@@ -0,0 +1,300 @@
+piercing-fragment-host[fragment-id="news"] {
```
dario-piotrowicz, line 1: Note: I'm disregarding the css here as I plan to tweak it afterwards (since @jculvey gave me the carte blanche 😍)

productivity-suite/app/fragments/news/src/lib/api.ts
```diff
@@ -0,0 +1,27 @@
+import { isServer } from "solid-js/web";
+
+const story = (path: string) => `https://node-hnapi.herokuapp.com/${path}`;
+const user = (path: string) =>
```
dario-piotrowicz, line 4: are we using this `user` bit?

productivity-suite/app/fragments/news/src/components/story.tsx
```diff
@@ -0,0 +1,59 @@
...
+          </a>
+          <span class="host"> ({props.story.domain})</span>
+        </Show>
+      </span>
+      <br />
+      <span class="meta">
+        <Show
+          when={props.story.type !== "job"}
+          fallback={
+            <A href={hnUrl(`/item?id=${props.story.id}`)}>
```
dario-piotrowicz, line 36: why these are `A` instead of `a` (like above)? do we still have some client side navigation? if not I guess we could also add the `target="_blank"`?

productivity-suite/app/fragments/news/src/root.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  let script: ManifestEntry = null;
+  let style: ManifestEntry = null;
+
+  if (isServer) {
+    const manifestEntries = context.env.manifest[
+      "entry-client"
+    ] as ManifestEntry[];
+    script = manifestEntries.find((e) => e.type === "script");
+    style = manifestEntries.find((e) => e.type === "style");
```
dario-piotrowicz, line 28: what does this stuff do? is it something solid forces you to do?

dario-piotrowicz (commented):
@jculvey sorry in my previous review I didn't check a couple of files! all reviewed now 🙇

productivity-suite/app/fragments/news/src/root.tsx
jculvey: Let's keep it around, it forces a reload in the HMR plugin: https://github.com/solidjs/solid-refresh#pragma

Personally I think it's be better to not getting rid of things you don't understand. I think it's much more likely we'd not notice something, especially if it's a subtle side effect, until much later, and then have a hard time backtracking from a working solution to figure out what changed.

productivity-suite/app/fragments/news/package.json
jculvey: it appears that this package.json is using 2-space and the one I copy pasted from uses tabs :(

productivity-suite/app/fragments/news/src/root.tsx
jculvey: `isSSR` is true when the app is in SSR mode (as determined by the config passed into the solid vite plugin), `isServer` is true when the code is running in the worker.

dario-piotrowicz: I am not sure I get it.... what's the difference? can't you just use one? 🤔

productivity-suite/app/fragments/news/src/root.tsx
jculvey: Nope, this is a hack. Solid normally handles this for you by automatically injecting the entry stylesheet and bundle when you use the `Head` and `Scripts` elements from `solid-start`. However, base path can't be provided as a param to either, so it doesn't work with fragments since we need to prepend `/_fragment/news`. 

Since the entry bundles are hashed, we have to lookup the dynamic filename via the manifest, which as passed down through context.

dario-piotrowicz: 😬 

could there be more then one script and style?

productivity-suite/app/fragments/news/src/root.tsx
jculvey: Nope, they're different. An app is in SSR mode or not. If you have an app running in SSR mode, the code can execute on either the server (as part of the initial async rendering on the worker) or on the client (as part of hydration). Similar to routing perhaps, you can have an SSR app that has routing, but your routing code will need to check if it's in a server context to know whether to use `document.location` or the request from context.

Also FYI these aren't variables I made up. They're copied from the `solid-start` source: https://github.com/solidjs/solid-start/blob/main/packages/start/root/Scripts.tsx

IgorMinar (changes requested):
Hey James, can you please add the news fragment to the top level package.json workspaces list? otherwise the deps don't install correctly when you run `npm install` at the root. Thank you.

productivity-suite/app/fragments/news/src/root.tsx
jculvey: In this particular example, no

productivity-suite/app/fragments/news/src/types.ts
jculvey: FWIW I too hate the `IWhatever` convention, but I didn't want to spend time making the code match our conventions for the sake of consistency. I'm guessing Pete's out for the day and I don't want to hold this over the weekend over a single character, so I don't mind changing it.

That being said, I think it's better to keep the solid example close to the source material because:
- If people are actually following along and comparing this repo to the source hackernews repo, how will people know if we've changed something because it's necessary for piercing vs because we don't like the conventions.
- If we decide to update the hackernews code later, we'll have to re-apply all the tweaks we made. If we're only deleting code (but not modifying it) this is much easier to do. 
- The goal was to take an existing app in another framework and have it work with piercing. We're trying to showcase piercing, not show people how to build a solid app according to our standards.
- Having multiple apps with varying conventions and quality levels simulates a more real world scenario.

productivity-suite/app/fragments/news/src/root.tsx
dario-piotrowicz: but in general yes?

I am asking because I think that in general we should not include things that can't handle generic situations and/or we can see a path forward for it

productivity-suite/app/fragments/news/src/root.tsx
dario-piotrowicz: ah.... I see.... the names are pretty confusing to me, but I sort of get it now

thanks for the clarification 🙂

productivity-suite/app/fragments/news/src/root.tsx
jculvey: Solid start implements it like this, so I think we're ok: 

`isSSR ? (
            <script type="module" async src={context.env.manifest["entry-client"][0].href} />`

dario-piotrowicz: ah 😮, ok thanks 🙂

dario-piotrowicz:
Locally when you run the application it seems like the news fragment gets rendered when it shouldn't:
Screenshot 2022-11-02 at 09 41 52

do you know why that is? and if we can fix it?

productivity-suite/app/fragments/news/vite.config.ts
```diff
@@ -0,0 +1,50 @@
...
+  ssr: {
+    // Vite attempts to load this as a Commonjs dependency
+    noExternal: ["solid-meta"],
+  },
+  // build: {
+  //   minify: false,
+  //   rollupOptions: {
+  //     external: ["__STATIC_CONTENT_MANIFEST"],
+  //   },
+  // },
```
dario-piotrowicz, line 42: is this bit wanted/needed?

productivity-suite/app/fragments/news/vite.config.ts
```diff
@@ -0,0 +1,50 @@
+import solid from "solid-start/vite";
+import cloudflareWorkers from "solid-start-cloudflare-workers";
+import { defineConfig, PluginOption } from "vite";
+
+export function addDefaultFnExportToBundle(regex: RegExp): PluginOption {
```
dario-piotrowicz, line 5: nit: I would maybe update the name of the function (like `wrapAndExportMount`), just to separate it from the other one? 🤔 (not a big deal, kinda just thinking out loud)

productivity-suite/app/legacy-app/src/worker/script.ts
```diff
@@ -173,4 +173,60 @@ gateway.registerFragment({
...
+  prePiercingStyles: `
+    :not(piercing-fragment-outlet) > piercing-fragment-host[fragment-id="news"] {
+      position: absolute;
+      margin: 0 1rem;
+      top: 13.9rem;
+      left: 0;
+      right: 0;
+    }
+
+    @media (max-width: 45rem) {
```
dario-piotrowicz, line 191: awesome, thanks for the positioning, I hope it wasn't too tedious 🙂 (I think this is quite one dx thing that it's not really the best here 😓)

productivity-suite/app/legacy-app/src/worker/script.ts
```diff
@@ -173,4 +173,60 @@ gateway.registerFragment({
...
+      :not(piercing-fragment-outlet) > piercing-fragment-host[fragment-id="news"] {
+        top: 29.65rem;
+        margin: 0;
+      }
+    }
+    `,
+  shouldBeIncluded: async (request: Request) =>
+    isPiercingEnabled(request) &&
+    (await isUserAuthenticated(request)) &&
+    /^\/(news(\/[^/]+)?)?$/.test(new URL(request.url).pathname),
```
dario-piotrowicz, line 213: This also matches `/` which shouldn't as that just redirects to the todos page
Screenshot 2022-11-02 at 09 41 52

dario-piotrowicz:
```suggestion
    /^\/news(\/[^/]+)?$/.test(new URL(request.url).pathname),
```

productivity-suite/app/legacy-app/src/worker/script.ts
```diff
@@ -173,4 +173,60 @@ gateway.registerFragment({
...
+    const url = new URL(request.url);
+    const path = url.pathname;
+    const match = /\/news\/([^/]+)$/.exec(path);
+    const listName = match?.[1] && decodeURIComponent(match[1]);
+
+    const params = url.searchParams;
+    if (listName) {
+      params.append("listName", listName);
+    }
+    return new Request(`${thisConfig.getBaseUrl(env)}?${params}`, request);
```
dario-piotrowicz, line 228: the lists stuff is actually not relevant here, I think you could replace it with a todo for your next change?
```suggestion
    // todo: add pagination and/or routing to request searchParams here
    return request;
```

productivity-suite/app/fragments/news/src/types.ts
petebacondarwin: Let's avoid changing the Solid example code if possible. I agree with the reasoning above.

petebacondarwin: I also do not believe in prefixing interfaces with `I`.

dario-piotrowicz: 🥲👍

productivity-suite/app/fragments/news/package.json
petebacondarwin: Let's add the prettifier to this branch and demo too...

dario-piotrowicz:
@jculvey I forgot to mention it, but Igor mentioned to have the news page next to the todos one in the nav, could you make that change as well 🙂

productivity-suite/app/legacy-app/src/worker/script.ts
jculvey: No prob! This was probably the most fun I had in this PR tbh :D

productivity-suite/app/fragments/news/package.json
jculvey: If it's ok I'll do that in a follow-up PR to avoid merge conflicts with Dario's message bug PR

dario-piotrowicz (approved):

productivity-suite/app/legacy-app/src/components/pages/News.tsx
```diff
@@ -1,11 +1,13 @@
...
+  // @ts-ignore
+  window.e = window._$HY = null;
 
-const newsIcon = (
-  <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48">
-    <path d="M11.3 39.4q-1.15 0-1.925-.775Q8.6 37.85 8.6 36.7V11.3q0-1.15.775-1.925Q10.15 8.6 11.3 8.6h19.65l8.45 8.45V36.7q0 1.15-.775 1.925-.775.775-1.925.775Zm0-1.1h25.4q.7 0 1.15-.45.45-.45.45-1.15V17.85h-8.15V9.7H11.3q-.7 0-1.15.45-.45.45-.45 1.15v25.4q0 .7.45 1.15.45.45 1.15.45Zm4-7.05h17.4v-1.1H15.3Zm0-13.4H24v-1.1h-8.7Zm0 6.7h17.4v-1.1H15.3ZM9.7 9.7v8.15V9.7 38.3 9.7Z" />
-  </svg>
-);
+  // @ts-ignore
+  window.t = function () {};
```
petebacondarwin, line 6: Oh I meant to say, we should add comments to the code for these workarounds. @jculvey could you do that in a follow-up PR?

PR #40, opened 2022-10-31
chore(cloud-gallery): prettier code!

petebacondarwin:


cloud-gallery/.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*]
+indent_style = tab
+indent_size = 2
+trim_trailing_whitespace = true
+insert_final_newline = true
```
dario-piotrowicz, line 7:
```suggestion
insert_final_newline = true

```

.github/workflows/test.yml
```diff
@@ -0,0 +1,30 @@
...
+          node-version: 16.13.x
+
+      - name: Install
+        working-directory: ./cloud-gallery
+        run: npm ci
+
+
+      - name: Run tests
+        working-directory: ./cloud-gallery
+        run: npm run -w helpers build && npm run test
```
dario-piotrowicz, line 30: not a big deal but I find `test` here pretty misleading since we don't actually have any proper testing

cloud-gallery/.prettierrc.json
```diff
@@ -0,0 +1 @@
+{}
```
dario-piotrowicz, line 1: no rules at all?

cloud-gallery/body/package.json
```diff
@@ -7,6 +7,8 @@
 		"build.client": "vite build",
 		"build.ssr": "vite build --ssr src/entry.ssr.tsx",
 		"lint": "eslint \"src/**/*.ts*\"",
+		"check-format": "prettier --check --ignore-path=../.prettierignore .",
+		"format": "prettier --write --ignore-path=../.prettierignore .",
```
dario-piotrowicz, line 11: is the `--ignore-path` part necessary? 😮

dario-piotrowicz (approved):
🤩

cloud-gallery/.editorconfig
petebacondarwin: LOL!

.github/workflows/test.yml
petebacondarwin: I'll change it to check

cloud-gallery/.prettierrc.json
petebacondarwin: They are all covered by `.editorconfig`.

cloud-gallery/body/package.json
petebacondarwin: Yes because we are running prettier in the `cloud-gallery/body` (for example) directory and by default prettier will only look for `./.prettierigore` (i.e. in the local directory).

cloud-gallery/.prettierrc.json
dario-piotrowicz: ah I had no idea that prettier would pick that up! 🙂 

in the [prettier docs](https://prettier.io/docs/en/configuration.html#editorconfig) they say that you need to add an option for that? is that not necessary here?

petebacondarwin: These docs suggest that this config option is on by default: https://prettier.io/docs/en/cli.html#--no-editorconfig

dario-piotrowicz: ah yeah I tried to google it earlier but it wasn't clear at all 😓:+1:

dario-piotrowicz (approved):

PR #44, opened 2022-11-04
Productivity suite - select latest list for user when they login

dario-piotrowicz:
Given our latest implementation when the user logs in they won't have a selected list, this causes the todos app to break.

So we need to make sure that when users are logged in a list is selected for them.

(Before we always had that list in the message bus but now we don't so we need to select it upon login)

### DIFF
--- a/productivity-suite/app/legacy-app/src/components/pages/Login.tsx
+++ b/productivity-suite/app/legacy-app/src/components/pages/Login.tsx
@@ -1,6 +1,7 @@
 import { getBus } from "piercing-library";
 import { useEffect, useRef } from "react";
 import { useLocation, useNavigate } from "react-router-dom";
+import { getTodoLists } from "shared";
 import { useAuth } from "../../auth";
 
 export function Login() {
@@ -16,10 +17,14 @@ export function Login() {
       return (
         getBus(ref.current).listen<{ username: string }>(
           "authentication",
-          (authDetails) => {
+          async (authDetails) => {
             if (authDetails?.username) {
-              auth.login(authDetails.username);
-              navigate(from, { replace: true });
+              await auth.login(authDetails.username);
+              const todoLists = await getTodoLists(authDetails.username);
+              getBus(ref.current!).dispatch("todo-list-selected", {
+                name: todoLists[todoLists.length - 1].name,
+              });
+              setTimeout(() => navigate(from, { replace: true }));
             }
           }
         ) ?? undefined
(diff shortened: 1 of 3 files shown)


productivity-suite/app/legacy-app/src/components/pages/Login.tsx
```diff
@@ -16,10 +17,14 @@ export function Login() {
...
+          async (authDetails) => {
             if (authDetails?.username) {
-              auth.login(authDetails.username);
-              navigate(from, { replace: true });
+              await auth.login(authDetails.username);
+              const todoLists = await getTodoLists(authDetails.username);
+              getBus(ref.current!).dispatch("todo-list-selected", {
+                name: todoLists[todoLists.length - 1].name,
+              });
+              setTimeout(() => navigate(from, { replace: true }));
```
dario-piotrowicz, line 27: unfortunately I needed to do this so that the navigation occurs after the `todo-list-selected`'s been properly set, otherwise the fragments would be fetched/added to the page too early

productivity-suite/app/legacy-app/src/components/pages/Login.tsx
```diff
@@ -16,10 +17,14 @@ export function Login() {
...
-          (authDetails) => {
+          async (authDetails) => {
             if (authDetails?.username) {
-              auth.login(authDetails.username);
-              navigate(from, { replace: true });
+              await auth.login(authDetails.username);
+              const todoLists = await getTodoLists(authDetails.username);
+              getBus(ref.current!).dispatch("todo-list-selected", {
+                name: todoLists[todoLists.length - 1].name,
+              });
```
petebacondarwin, line 26: This doesn't feel like the right solution here.
The Login page shouldn't know about the Todos page.
Can't we fix the Todos page so that it works this out for itself?

petebacondarwin (changes requested):

productivity-suite/app/legacy-app/src/components/pages/Login.tsx
dario-piotrowicz: Yeah I agree with you, if we don't want the login functionality to select the list for the user then I think we may need to refactor the todo lists selection logic we applied to fix the navigation issues

I am not sure if the current solution can handle it or if we may need to refactor back to the previous implementation and try to find a different solution for the navigation issues

what do you think?

petebacondarwin: Let's talk on Monday about this. I have not looked into it but we should definitely not be in a situation where a view cannot initialize itself.

dario-piotrowicz: I've applied the workaround we've discussed about, hopefully it's ok for now 🙂

petebacondarwin (approved):
LGTM. We can follow up with how to manage URL changes better.

PR #46, opened 2022-11-06
Productivity Suite - minor fixes related to news fragment

dario-piotrowicz:


productivity-suite/app/legacy-app/index.html
```diff
@@ -223,6 +224,10 @@
         content: 'login fragment';
       }
 
+      body.show-seams piercing-fragment-host[fragment-id="news"]::before {
+        content: 'news fragment';
```
dario-piotrowicz, line 228: before:
Screenshot 2022-11-06 at 18 41 18

after:
Screenshot 2022-11-06 at 18 41 33

productivity-suite/app/legacy-app/index.html
```diff
@@ -161,6 +161,7 @@
       }
 
       piercing-fragment-host {
+        display: block;
```
dario-piotrowicz, line 164: I've moved the display block here because it needs to always apply, otherwise we have different widths for the fragments depending on whether seams are on or not

petebacondarwin (approved):

PR #50, opened 2022-11-07
Productivity suite - parallelize fetches

dario-piotrowicz:


productivity-suite/piercing-library/src/piercing-gateway.ts
```diff
@@ -100,14 +99,6 @@ export class PiercingGateway<Env> {
     this.fragmentConfigs.set(fragmentConfig.fragmentId, fragmentConfig);
   }
 
-  /*
-		We're assigning fetch to the object itself so that user can do:
-			`export default gateway;`
-		this shouldn't be necessary and we should be able to have fetch
-		as a normal method in the class, but that doesn't seem to
-		currently being recognized correctly:
-			https://chat.google.com/room/AAAADFUqSgo/njWMATzSHG4
```
dario-piotrowicz, line 109: this refers to a private gchat conversation so I think we can/should just exclude it (it is also not an issue related to the piercing strategy)

petebacondarwin: Actually, I think the main comment is a valid thing to keep here. Perhaps just remove the link to the gchat?

petebacondarwin (approved):

productivity-suite/piercing-library/src/piercing-gateway.ts
```diff
@@ -3,7 +3,6 @@ import {
   piercingFragmentHostInlineScript,
 } from "./index";
 import { concatenateStreams, wrapStreamInText } from "./stream-utilities";
-// import qwikloader from "@builder.io/qwik/qwikloader.debug.js?raw";
```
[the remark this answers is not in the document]
dario-piotrowicz: ah sorry I thought it was left here by mistake, I don't mind putting it back, maybe I can add a comment?

or could we just replace it with something like:
```ts
// for debugging replace `qwikloader.js` with `qwikloader.debug.js` to have the code non-minified 
```
?

productivity-suite/piercing-library/src/piercing-gateway.ts
dario-piotrowicz: I've put the comment back 🙂

petebacondarwin (approved):

PR #55, opened 2022-11-08
Productivity Suite - Small improvements

dario-piotrowicz:


productivity-suite/app/legacy-app/src/components/pages/Login.css
```diff
@@ -0,0 +1,4 @@
+.login-page {
```
dario-piotrowicz, line 1: I made some small tweaks so that we can have a nicer seams outline for the login fragment

before:
Screenshot 2022-11-08 at 18 39 42

after:
Screenshot 2022-11-08 at 18 39 15

petebacondarwin (approved):
Nice cleanup

PR #66, opened 2022-11-09
Productivity Suite  - small tweaks/improvements

dario-piotrowicz:


productivity-suite/piercing-library/src/piercing-gateway.ts
```diff
@@ -135,30 +144,33 @@ export class PiercingGateway<Env> {
...
-      const baseUrl = this.config.getBaseAppUrl(env).replace(/\/$/, "");
+      const baseUrl = this.config.getLegacyAppBaseUrl(env).replace(/\/$/, "");
       const indexBodyResponse = this.fetchBaseIndexHtml(
         new Request(baseUrl, request)
       ).then((response) => response.text());
 
+      const piercingEnabled =
+        !this.config.shouldPiercingBeEnabled ||
+        (await this.config.shouldPiercingBeEnabled(request, env, ctx));
+
```
petebacondarwin, line 155: Given that we are using `!piercingEnabled` below how about we just do:

```suggestion
      const piercingDisabled = await this.config.shouldPiercingBeEnabled.?(request, env, ctx);
```

productivity-suite/piercing-library/src/piercing-gateway.ts
```diff
@@ -135,30 +144,33 @@ export class PiercingGateway<Env> {
...
+                ctx
+              );
+
+              return new Promise(async (resolve) =>
+                resolve(await shouldBeIncluded)
+              ).then((shouldBeIncluded) =>
+                shouldBeIncluded
+                  ? this.fetchSSRedFragment(env, fragmentConfig, request)
+                  : null
+              );
```
petebacondarwin, line 172: How about:

```ts
        : Array.from(this.fragmentConfigs.values()).map(
              async (fragmentConfig) => {
                const shouldBeIncluded = await fragmentConfig.shouldBeIncluded(
                  request,
                  env,
                  ctx
                );

                return shouldBeIncluded
                  ? this.fetchSSRedFragment(env, fragmentConfig, request)
                  : null;
              }
            );

```

petebacondarwin (approved):
Other than my two NITs this LGTM

productivity-suite/piercing-library/src/piercing-gateway.ts
dario-piotrowicz: is this not missing a negation?

if `this.config.shouldPiercingBeEnabled(request, env, ctx)` returns `true`, we get `piercingDisabled === true` which is the opposite of what we want

no?

productivity-suite/piercing-library/src/piercing-gateway.ts
dario-piotrowicz: ah yeah, I was thinking of something like this but I was wondering if it could mess up the order of the promises (I played a bit with it and it doesn't seem to)

anyways the order wouldn't be that important either way :+1:

productivity-suite/piercing-library/src/piercing-gateway.ts
petebacondarwin: Doh! Yeah sorry. Ignore this.

petebacondarwin (approved):

PR #67, opened 2022-11-10
Update productivity start dev script

petebacondarwin:


productivity-suite/scripts/start-dev.js
```diff
@@ -37,31 +37,26 @@ async function runAllProcesses() {
...
 	}
 
-    await sleep(3000);
+    log('Serving the legacy application...');
+    runNpmScript(path.join(appDir, 'legacy-app'), 'dev.react');
+
+    await sleep(2000);
+
+    log(`Serving the Gateway worker...`);
+    runNpmScript(path.join(appDir, 'legacy-app'), 'dev.worker');
```
dario-piotrowicz, line 56: are you sure this is ok? I remember specifically starting the gateway worker first because otherwise it seemed like it would not recognise the bindings (or something like that) 🤔

dario-piotrowicz: I've just tested it locally and it seems to work fine... maybe something has changed... or maybe I did something wrong back then 🤷

petebacondarwin: It should not matter too much since the way Wrangler dev mode works, it proxies the service-bindings anyway. But... in general, it is safer to run the dependencies first. In the case of deployment (or remote dev mode) you absolutely need to have the dependencies uploaded before the depending workers.

PR #68, opened 2022-11-11
fix(productivity-suite): split the auth messages into three

petebacondarwin:
Previously, we were using a single MessageBus message (authentication) to handle logging in, logging out, and the current state of authentication. This caused lots of problems where the React router was triggering redirects while the authentication state was still loading.

Now we have three MessageBus messages:

- login - the user has clicked a login button
- logout - the user has clicked a logout button
- authtentication - the current auth state of the user


productivity-suite/app/legacy-app/src/components/pages/UnimplementedDemoPage.tsx
```diff
@@ -18,8 +19,10 @@ export function UnimplementedDemoPage({
         <p>This page is not implemented as it is out of scope for this demo.</p>
         <p>
           To Interact with SSR fragments view the login form by{" "}
-          <button onClick={auth.logout}>logging out</button> or visit the{" "}
-          <Link to="/todos">todos page</Link>.
+          <button onClick={() => getBus().dispatch("logout", null)}>
```
dario-piotrowicz, line 22: should we consider making the second argument of `dispatch` optional for such type of events?

productivity-suite/app/legacy-app/src/auth.tsx
```diff
@@ -3,44 +3,52 @@ import { createContext, useContext, useEffect, useState } from "react";
...
+    AUTH_LOADING
+  );
 
-  const [user, setUser] = useState<string | null>(null);
   useEffect(() => {
-    return getBus().listen<{ username: string }>("authentication", (auth) => {
-      setUser(auth?.username);
-    });
-  });
+    getCurrentUser().then((username) => setUser(username));
```
dario-piotrowicz, line 27: would this look nicer with async/await?

```ts
useEffect(async () => setUser(await getCurrentUser()));
```

dario-piotrowicz: this is getting the user from the cookie and ignoring the one present in the message bus

I think we should get the data from the message bus, shouldn't we?

(
and if we don't I think we should clean this up?
https://github.com/cloudflare/workers-web-experiments/blob/56a556a31d03428ea503721d3636b8dc0aa38487/productivity-suite/app/legacy-app/src/worker/script.ts#L25-L29
)

productivity-suite/app/legacy-app/src/auth.tsx
```diff
@@ -3,44 +3,52 @@ import { createContext, useContext, useEffect, useState } from "react";
...
 
-interface AuthContextType {
-  user: string | null;
-  login: (userName: string) => void;
-  logout: () => void;
-}
+export const AUTH_LOADING = Symbol("Auth Loading");
+export type AuthContextType = { user: string | null } | undefined;
 
 let AuthContext = createContext<AuthContextType>(null!);
```
dario-piotrowicz, line 16: with the current changes is `null!` still needed?

productivity-suite/app/legacy-app/src/auth.tsx
```diff
@@ -3,44 +3,52 @@ import { createContext, useContext, useEffect, useState } from "react";
...
 
 export function AuthProvider({ children }: { children: React.ReactNode }) {
   const navigate = useNavigate();
+  const location = useLocation();
+
+  const [user, setUser] = useState<string | null | typeof AUTH_LOADING>(
+    AUTH_LOADING
+  );
```
dario-piotrowicz, line 24: I am not sold on this `AUTH_LOADING` 🤔 

if I understand correctly the code, we have the user data which we wrap and unwrap
(for example here:
```ts
const value = user !== AUTH_LOADING ? { user } : undefined;
```
)
, but couldn't we just use the object itself?
(for example:
```ts
const [auth, setAuth] = useState<AuthContextType>();

// ....

const value = auth;
```
)

do we get anything out of the `AUTH_LOADING`?

productivity-suite/app/legacy-app/src/auth.tsx
dario-piotrowicz: ah, I think it is, but it should be:
```suggestion
let AuthContext = createContext<AuthContextType>(undefined);
```
I think?

productivity-suite/app/legacy-app/src/auth.tsx
jculvey: The proposed code is similar to what we had before. I think the issue we discovered is that using just a single object allows for 2 states: 1) object exists and user is therefore authed, or 2) object is null and therefore isn't authed.

But in reality we have 3 states we need to handle: 1) initializing 2)authed, 3) not-authed (but initialized). This initializing state is important so that we can distinguish between needing to redirect after a logout (when the object would be null) but not immediately after login (when the object would also be null).

productivity-suite/app/legacy-app/src/auth.tsx
jculvey: Can't pass async functions to `useEffect`

productivity-suite/app/legacy-app/src/components/pages/UnimplementedDemoPage.tsx
jculvey: I like the sound of that

jculvey (approved):

productivity-suite/app/legacy-app/src/auth.tsx
jculvey: You can't pass `async` functions to `useEffect`

productivity-suite/app/legacy-app/src/auth.tsx
```diff
@@ -62,7 +71,8 @@ export function RequireAuth({ children }: { children: React.ReactNode }) {
 export function RequireNotAuth({ children }: { children: React.ReactNode }) {
   const auth = useAuth();
 
-  if (auth.user) {
+  // Don't redirect if the auth has not yet been initialized.
+  if (auth && auth.user) {
```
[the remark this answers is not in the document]
petebacondarwin: I'm going to push back on this suggestion. The comment is there to explain an aspect of the code that is not so obvious, which is that `auth` being undefined means it is not yet initialized and we don't want to do anything yet in that case. The rest of the code is self-explanatory and so doesn't need a further comment.

productivity-suite/app/legacy-app/src/components/pages/UnimplementedDemoPage.tsx
petebacondarwin: Yeah I was wondering about that too.
I think that is OK but in such cases what happens when we are serializing the event? We currently serialize to JSON the map of event names to values, but I don't think we can serialize `{ myEvent: undefined }` without losing the event name altogether.

productivity-suite/app/legacy-app/src/auth.tsx
petebacondarwin: I think so. Thanks

productivity-suite/app/legacy-app/src/auth.tsx
dario-piotrowicz: ahhh, ok sorry my bad 😅

productivity-suite/app/legacy-app/src/auth.tsx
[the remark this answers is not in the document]
dario-piotrowicz: in this piece of code you are checking if `auth.user` is truthy... and it if is you are doing the redirect

so it is very confusing to me that this is saying not to do something right before the code that does that thing... (even though there is an if check)

to be honest I thought it was a copy/paste mistake

I think that this code would actually be clearest without any comment at all

but I don't want to insist, if you believe this is better I'm ok with it

productivity-suite/app/legacy-app/src/components/pages/UnimplementedDemoPage.tsx
dario-piotrowicz: what I was thinking was to actually have the second argument optional and it could default to `null`:
```ts
  dispatch(eventName: string, value: JSONValue = null) 
```
so it wouldn't really change anything, just provide a shorthand to dispatch a `null` value

dario-piotrowicz: PS: I just noticed that we've got an `any` there instead of `JSONValue` 😅 

Screenshot 2022-11-11 at 21 32 07

productivity-suite/app/legacy-app/src/auth.tsx
dario-piotrowicz: I am not sure I get that, it looks to me like the three states are handled like so from the outside of the AuthContext (and I am just suggesting to also apply the same logic inside the AuthContext so that there there aren't two different representation of the same information, the extra symbol is not necessary and there is no extra wrapping/unwrapping):

1) initializing, authContext is undefined
2) authed, authContext is { user: string }
3) not-authed (but initialized), authContext is { user: null }

and to me it actually makes a lot of sense saying that when the context is undefined it means that it is initializing/the information is not yet there, then when it is defined it can tell us if the user is authed or not

or am I missing something?

productivity-suite/app/legacy-app/src/auth.tsx
dario-piotrowicz: by the way, I still stand with my comment that I think the data should come from the bus 😜

productivity-suite/app/legacy-app/src/auth.tsx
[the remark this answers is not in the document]
petebacondarwin: OK. I changed the comments. Hopefully they are more clear. Thanks for picking me up on this!

productivity-suite/app/legacy-app/src/auth.tsx
petebacondarwin: You are right, we can indeed just use the message bus to get the initial value.
This messed with my brain when we were working on this because I couldn't work out where this would be initialized.
But I think it is always done in the Worker, right?

Anyway, I changed it and all seems well 🙏

productivity-suite/app/legacy-app/src/auth.tsx
petebacondarwin: You convinced me! I updated the code. Thanks

productivity-suite/app/legacy-app/src/auth.tsx
dario-piotrowicz: Thanks 😄 

Yeah it should be all done by the piercing gateway, at the beginning of the request handling we set the value using the user provided `generateMessageBusState`:

https://github.com/cloudflare/workers-web-experiments/blob/0da0c31dfce27e8ae5c992cb57565efddd106cd6/productivity-suite/app/legacy-app/src/worker/script.ts#L21-L29

(and the message bus is applied by the piercing gateway as it fetches the index.html:
https://github.com/cloudflare/workers-web-experiments/blob/0da0c31dfce27e8ae5c992cb57565efddd106cd6/productivity-suite/piercing-library/src/piercing-gateway.ts#L310)

productivity-suite/app/legacy-app/src/auth.tsx
dario-piotrowicz: awesome thanks! 😃

productivity-suite/app/legacy-app/src/auth.tsx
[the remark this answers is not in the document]
dario-piotrowicz: Thanks 😃

dario-piotrowicz (approved):

productivity-suite/app/legacy-app/src/auth.tsx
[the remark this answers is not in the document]
dario-piotrowicz: I'm not sure if I'm overthinking this (I probably am! 😝)

what if defined a (file-local) function like:
```ts
function isAuthInitialized(auth: AuthContext) {
   return !!auth;
}
```
?

then we could have:
```diff
-  // If `auth` is undefined then it has not yet been initialized
-  // so we should avoid making any premature redirect navigation.
-  if (auth && !auth.username) {
+ if (isAuthInitialized(auth) && !auth.username) {
    return <Navigate to={"/login"} replace />;
  }

// ...

-  // If `auth` is undefined then it has not yet been initialized
-  // so we should avoid making any premature redirect navigation.
-  if (auth && auth.username) {
+ if (isAuthInitialized(auth) && auth.username) {
    return <Navigate to={"/"} replace />;
  }
  ```
  
such code looks to me very clear to understand so we wouldn't need to add any comment at all

what do you think? 🙂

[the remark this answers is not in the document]
dario-piotrowicz: (PS: but this is not a blocker or anything, just an idea 😄)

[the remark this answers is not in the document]
petebacondarwin: Actually now that we are initializing the auth synchronously from the message bus, we can be sure that this is never undefined by the time we use it. I will update to add an assertion just to be sure.

PR #71, opened 2022-11-14
fix(productivity-suite): apply visibility hidden fix

dario-piotrowicz:
make the legacy application not visible until all the initial page fragments have been pierced (this helps solving the issue of the footer quickly flashing)

### DIFF
--- a/productivity-suite/piercing-library/src/piercing-fragment-host/piercing-fragment-host.ts
+++ b/productivity-suite/piercing-library/src/piercing-fragment-host/piercing-fragment-host.ts
@@ -29,7 +29,7 @@ export class PiercingFragmentHost extends HTMLElement {
 
     if (!this.fragmentIsPierced) {
       this.setStylesEmbeddingObserver();
-      updateFragmentsToPierce("++");
+      increaseFragmentsToPierceCount();
     }
   }
 
@@ -52,7 +52,7 @@ export class PiercingFragmentHost extends HTMLElement {
     this.cleanup = true;
 
     activeElement?.focus();
-    updateFragmentsToPierce("--");
+    decreaseFragmentsToPierceCount();
   }
 
   onPiercingComplete() {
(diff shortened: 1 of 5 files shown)


productivity-suite/piercing-library/src/piercing-fragment-outlet.ts
```diff
@@ -82,6 +82,9 @@ export class PiercingFragmentOutlet extends HTMLElement {
   private pierceFragmentIntoDOM(fragmentHost: PiercingFragmentHost) {
     this.innerHTML == "";
     fragmentHost.pierceInto(this);
+    const fragmentsToPierce =
+      getBus().latestValue<number>("fragmentsToPierce")!;
+    getBus().dispatch("fragmentsToPierce", fragmentsToPierce - 1);
```
petebacondarwin, line 87: Why not do this inside the `pierceInto()` method, since there is already access to the message bus there.

petebacondarwin (changes requested):
I love the use of the message bus to handle the count of fragments.
I think we could avoid touching both the outlet and the gateway by doing this in the fragment-host:

- in the `connectedCallback()`, we already have logic to check if the fragment is pierced and then add the piercing styles. How about we increment the fragment count there?
- then in the `pierceInto()` method we can decrement the fragment count.

(Also, not part of this PR - I see that we no longer call `removeStylesEmbeddingObserver()`, since we never call `onPiercingComplete()`. I think that we should do this in the `pierceInto()` method?)

productivity-suite/piercing-library/src/piercing-fragment-outlet.ts
dario-piotrowicz: ah yeah.... good point 😓

dario-piotrowicz:
Thanks for the comment @petebacondarwin, it felt natural to me to do the thing in the gateway, but I guess it does make more sense to just do it in the fragment host 🙂, I'll make the change right away 👍

productivity-suite/piercing-library/src/piercing-fragment-host/piercing-fragment-host.ts
```diff
@@ -102,3 +105,13 @@ export class PiercingFragmentHost extends HTMLElement {
     });
   }
 }
+
+function updateFragmentsToPierce(operation: "++" | "--") {
```
dario-piotrowicz, line 109: @petebacondarwin, what do you think about the operation argument there?

I sort of wonder if it is neat or awkward 😅

dario-piotrowicz:
This change seems no longer necessary after https://github.com/cloudflare/workers-web-experiments/pull/68

PR #72, opened 2022-11-14
Productivity Suite - styles and mobile improvements

dario-piotrowicz:
In this PR I've applied fixes that should hopefully make the app better on iPhones

besides that I replaced the nav on small screens with a select-like navigation:
nav-list

### DIFF
--- a/productivity-suite/app/fragments/todo-lists/src/components/TodoListsCarousel.css
+++ b/productivity-suite/app/fragments/todo-lists/src/components/TodoListsCarousel.css
@@ -57,6 +57,7 @@
     align-self: center;
     font-family: inherit;
     font-weight: 300;
+    padding: 0.1rem;
 }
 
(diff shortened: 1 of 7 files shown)


petebacondarwin (approved):
I hate to say it but there is a bit of movement on the mobile views in my local Chrome browser 😓  I suspect it is only me so I won't block this landing. But can you double check?

dario-piotrowicz:
> I hate to say it but there is a bit of movement on the mobile views in my local Chrome browser 😓 I suspect it is only me so I won't block this landing. But can you double check?

ah, sorry about that 😢 

on mobile I can also see a small vertical shift of the fragments, I thought I did check that... sorry about that I'll fix it in a follow up PR (I'll try to make it as precise as possible for my setup and for medium font sizes and hopefully this will create a generally acceptable effect, does that sound ok?)

PR #74, opened 2022-11-14
feat(productivity-suite): make news page persist

dario-piotrowicz:
use search parameters to update and restore the news page so that they can persist across refreshes and the user can also see that not only the first page can get server side rendered


dario-piotrowicz:
It's just an idea, let me know what you think 🙂

productivity-suite/app/legacy-app/src/components/pages/News.tsx
```diff
@@ -11,8 +15,20 @@ export function News() {
     };
   });
 
+  const ref = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (ref.current) {
+      return getBus(ref.current).listen<number>("news-page-set", (newsPage) => {
```
petebacondarwin, line 22: Since this is not inside a fragment, you don't actually need a `ref` you can just call `getBus()`. But perhaps that might be confusing?

productivity-suite/app/fragments/news/src/routes/[...news].tsx
```diff
@@ -1,18 +1,30 @@
...
+import { useSearchParams } from "solid-start";
 import Story from "~/components/story";
 import fetchAPI from "~/lib/api";
 import { IStory } from "~/types";
+import { getBus } from "piercing-library";
 
 export default function News() {
   const [page, setPage] = createSignal<number>(1);
 
+  const [searchParams] = useSearchParams<{ page?: string }>();
```
Comment (petebacondarwin) [inline productivity-suite/app/fragments/news/src/routes/[...news].tsx:19]: Should the fragment be reading the search params?
I feel like it should be reading and writing this on the message bus?

productivity-suite/app/legacy-app/src/components/pages/News.tsx
dario-piotrowicz: yes.... the fact is that you could never use the ref/element argument and everything would still work fine (except from memory leaks)

so I was sort of thinking that maybe we should always try to use an element whenever possible and only use `getBus()` when there are no elements, that does sound like a sort best-practice-like/of clear cut rule... but I am not sure if that makes too much sense

otherwise I guess the rule here would be to never provide the element and only do so inside fragments?
(PS: but also there, if you only dispatch a value you could as well just not use the element right?)

productivity-suite/app/fragments/news/src/routes/[...news].tsx
Comment (dario-piotrowicz) [reply productivity-suite/app/fragments/news/src/routes/[...news].tsx:19]: yeah I fell in the same mistake I did for the todos 🤦 

sorry about that

productivity-suite/app/legacy-app/src/components/pages/News.tsx
```diff
@@ -11,9 +16,57 @@ export function News() {
     };
   });
 
+  useEffect(() => {
+    if (currentPageRef.current === undefined) {
+      const pageDetails = getBus().latestValue<{ page: number }>(
+        "update-news-page-number"
```
petebacondarwin, line 22: `update-new-page-number` and `update-news-page-view` seem very similar to me and it is difficult to tell which does what. Could we make them less similar?

Could it be simply: `news-page` and `change-news-page`?

dario-piotrowicz:
Sorry @petebacondarwin I had some issues with rebasing after we updated the main branch, so I re-created the PR (https://github.com/cloudflare/workers-web-experiments/pull/80) I hope you don't mind 🙇

PR #75, opened 2022-11-15
fix(productivity-suite): improve layout footer and todos page for SEO scores

petebacondarwin:
We were dropping a percent on the Lighthouse SEO score because
there was not enough room to tap on some of the links and buttons.
Lighthouse requires that tap-able areas are at least 48x48 in size
or away from other links.

### DIFF
--- a/productivity-suite/app/fragments/todo-lists/src/components/SelectedListCard.css
+++ b/productivity-suite/app/fragments/todo-lists/src/components/SelectedListCard.css
@@ -46,12 +46,23 @@
 }
 
 .delete-btn {
-    font-family: inherit;
-    font-size: 1.1rem;
     position: absolute;
-    top: 1rem;
-    right: 1rem;
+    top: -1.5rem;
+    right: -1.5rem;
+    padding: 1.5rem;
     z-index: inherit;
+    font-size: 2rem;
+    line-height: 1.1rem;
+    box-shadow: -3px 3px 5px #4d4d4d47, 0 0 20px #4d4d4d0d;
+    border-color: transparent;
+    border-radius: 2.4rem;
+    color: #4d4d4d;
+    background-color: #fff;
+    font-weight: 300;
+}
+
+.delete-btn:hover {
+    outline: solid 1px red;
 }
 
 .delete-btn:disabled {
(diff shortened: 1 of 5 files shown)


dario-piotrowicz (approved):
The delete button seems way too big (and distracting) now 😓, could it be smaller?

petebacondarwin:
This was the smallest I could make it that would still pass the SEO test 😢 
We could use this style only on mobile views?

dario-piotrowicz:
> This was the smallest I could make it that would still pass the SEO test 😢 We could use this style only on mobile views?

that's interesting, considering that now it is the biggest button on the whole page 😮

petebacondarwin:
I managed to squeeze it down a bit to this:
image

petebacondarwin:
It's all about the overlap with other buttons. If it was sitting in a load of empty space then it could be tiny, since the phone would have no problem guessing that you meant to click it.

dario-piotrowicz:
> It's all about the overlap with other buttons. If it was sitting in a load of empty space then it could be tiny, since the phone would have no problem guessing that you meant to click it.

yeah I was looking at the same thing, for example (a quick and ugly one 😅):
Screenshot 2022-11-15 at 14 54 45

In that case I think we should try to keep it more or less as small as before but make sure it doesn't overlap with the button text (on small screens if that's the only problem)

petebacondarwin:
Actually I have a better idea. There is no need for the user to see the extent of the button. We can make it transparent...

image

productivity-suite/app/fragments/todo-lists/src/components/SelectedListCard.css
```diff
@@ -46,12 +46,17 @@
...
-    font-size: 1.1rem;
     position: absolute;
-    top: 1rem;
-    right: 1rem;
+    top: -0.3em;
+    right: -0.3em;
+    padding: 1em;
     z-index: inherit;
+    font-size: 1.5rem;
+    line-height: 1ex;
```
dario-piotrowicz, line 55: interesting unit! 😃

dario-piotrowicz (approved):

PR #78, opened 2022-11-15
Productivity suite - style improvements

dario-piotrowicz:
Seams before and after:
Screenshot 2022-11-15 at 18 49 59

Screenshot 2022-11-15 at 18 50 25


productivity-suite/app/legacy-app/src/components/layout.css
```diff
@@ -32,6 +32,10 @@
   text-align: center;
 }
 
+.app-nav-list > li:has(a.active) {
+  margin-block: 0;
```
petebacondarwin, line 36: Out of interest, why margin-block rather than margin?

dario-piotrowicz: because with `margin: 0` we'd also be messing with the width of the elements and that would create some ugly movements, like so:
margin-block

productivity-suite/app/legacy-app/index.html
```diff
@@ -178,6 +178,13 @@
         max-width: min(calc(55rem + 2rem), calc(90vw));
       }
 
+    :not(piercing-fragment-outlet) > piercing-fragment-host:last-of-type::after {
+      content: '';
+      position: absolute;
```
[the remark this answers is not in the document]
dario-piotrowicz: yes, but we aren't actually expanding the container but adding something below it (so that it doesn't touches the bottom of the screen)

for example:

Screenshot 2022-11-15 at 18 38 35

Screenshot 2022-11-15 at 18 38 43

as you can see we're adding an extra bit there outside the container so that it "pushes down" the bottom of the screen

this works but it is not very elegant... if you'd got an alternative idea in mind that's be awesome 😃

productivity-suite/app/legacy-app/src/worker/script.ts
```diff
@@ -203,7 +203,7 @@ gateway.registerFragment({
 
     @media (max-width: 45rem) {
       :not(piercing-fragment-outlet) > piercing-fragment-host[fragment-id="news"] {
-        top: 14.81rem;
+        top: 13.835rem;
```
petebacondarwin, line 206: 😓

petebacondarwin (approved):
I see the tiniest movement on the todos and news pages but not worth trying to fix, given that it is probably just me.

productivity-suite/app/legacy-app/src/worker/script.ts
dario-piotrowicz: 😢

PR #80, opened 2022-11-16
feat(productivity-suite): make news page persist

dario-piotrowicz:
use search parameters to update and restore the news page so that they can persist across refreshes and the user can also see that not only the first page can get server side rendered


dario-piotrowicz:
@petebacondarwin regarding your comment:
> update-new-page-number and update-news-page-view seem very similar to me and it is difficult to tell which does what. Could we make them less similar?
>
> Could it be simply: news-page and change-news-page?

I've updated the names, I am not sure if in your message you were suggesting that `update-news-page-number` should be `news-page` and `update-news-page-view` should be `change-news-page`, but I did it the other way around:
`update-news-page-number` -> `change-news-page`
`update-news-page-view` -> `news-page`
because the standard one being read on the server and on the initial page load is the `update-news-page-view` whilst the other is only fired from the fragment to update the former, so I hope this makes sense

please have a look 🙂

dario-piotrowicz:
> > @petebacondarwin regarding your comment:
> > > update-new-page-number and update-news-page-view seem very similar to me and it is difficult to tell which does what. Could we make them less similar?
> > > Could it be simply: news-page and change-news-page?
> > 
> > 
> > I've updated the names, I am not sure if in your message you were suggesting that `update-news-page-number` should be `news-page` and `update-news-page-view` should be `change-news-page`, but I did it the other way around: `update-news-page-number` -> `change-news-page` `update-news-page-view` -> `news-page` because the standard one being read on the server and on the initial page load is the `update-news-page-view` whilst the other is only fired from the fragment to update the former, so I hope this makes sense
> > please have a look 🙂
> 
> LOL! This was exactly my problem. I couldn't tell which was which! Thanks for updating.

Yeah for me `-number` was to indicate the number of the page and `-view` the actual page that should be displayed... so basically the number of the page... 😅 

yeah sorry, pretty bad naming on my part 😓

petebacondarwin (commented):
Two quick comments but also I propose we change the clean up logic for fragments in general... See my extra commit.

productivity-suite/app/fragments/news/src/routes/[...news].tsx
```diff
@@ -16,10 +17,13 @@ export default function News() {
...
-    if (isFinite(page) && page > 0) {
-      setPage(page);
-    }
+  onMount(() => {
+    getBus(ref).listen<{ page: number }>("news-page", ({ page }) => {
+      if (isFinite(page) && page > 0) {
+        setPage(page);
+      }
+    });
+    getBus(ref).onCleanup(() => getBus(ref).dispatch("change-news-page", {}));
```
Comment (petebacondarwin) [inline productivity-suite/app/fragments/news/src/routes/[...news].tsx:26]: Here we add a handler that will be called when the fragment itself is being removed.

productivity-suite/app/legacy-app/src/components/pages/News.tsx
```diff
@@ -45,9 +24,11 @@ export function News() {
 
   useEffect(() => {
     return getBus().listen<{ page: number }>("change-news-page", ({ page }) => {
-      setSearchParams({
-        page: `${page}`,
-      });
+      if (page) {
```
petebacondarwin, line 27: Avoid setting the search params if there is not a valid page.
Otherwise you see things like:  `/calendar?page=undefined`

productivity-suite/piercing-library/src/message-bus/get-bus.ts
```diff
@@ -9,6 +10,8 @@ import { messageBusProp } from "./message-bus-prop";
  *                the message bus in the global scope is returned.
  * @returns       The most relevant message bus.
  */
+export function getBus(): MessageBus;
+export function getBus(element: Element): FragmentMessageBus;
```
petebacondarwin, line 14: This overload means that we get access to the `onCleanup()` function only when we pass a ref.
This is not strictly asserted at runtime right now... E.g. if you pass a `ref` outside of a fragment you get a `GenericMessageBus` rather than a `FragmentMessageBus`. But we could add a check for that somehow.

petebacondarwin: I guess we could add the `onCleanup()` method to the `GenericMessageBus` too but make it a noop?

productivity-suite/app/legacy-app/src/components/pages/News.tsx
```diff
@@ -11,9 +16,54 @@ export function News() {
...
 }
+
+function getPageNumberFromSearchParams(searchParams: URLSearchParams) {
+  const searchParamPage = searchParams.get("page");
+  const searchParamPageNum = searchParamPage ? +searchParamPage : 1;
+  const newsPageNum =
+    isFinite(searchParamPageNum) && searchParamPageNum > 0
+      ? searchParamPageNum
+      : 1;
+  return newsPageNum;
```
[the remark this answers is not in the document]
dario-piotrowicz: the problem is that search parameters can be defined directly by the user, so that's why I wanted to put extra checks in place

for example in your code, I think that if we had `?page=hello`, we would end up with `NaN` as the `newsPageNum` instead of gracefully falling back to `1`

productivity-suite/piercing-library/src/message-bus/get-bus.ts
dario-piotrowicz: I guess... would there be a benefit in that?

productivity-suite/app/fragments/news/src/routes/[...news].tsx
```diff
@@ -9,17 +9,31 @@ import {
...
 import fetchAPI from "~/lib/api";
 import { IStory } from "~/types";
+import { getBus } from "piercing-library";
 
 export default function News() {
+  let ref: HTMLDivElement;
+
   const [page, setPage] = createSignal<number>(1);
 
+  getBus(ref).listen<{ page: number }>("news-page", ({ page }) => {
```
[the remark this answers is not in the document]
Comment (dario-piotrowicz) [reply productivity-suite/app/fragments/news/src/routes/[...news].tsx]: yeah you're right, sorry, silly mistake 😓

productivity-suite/piercing-library/src/message-bus/get-bus.ts
petebacondarwin: It would mean that we could avoid these overloads here, and also the user wouldn't need to care about whether it is inside a fragment or not...

productivity-suite/app/legacy-app/src/components/pages/News.tsx
petebacondarwin: When we cleanup the fragment we need a way to specify that there is no longer any "news page". I don't believe that setting it to 1 when we have left the fragment is a good idea. So we must "wipe out" the message by dispatching `undefined` or `{page: undefined}`. Either way we would then need to handle the undefined case.

productivity-suite/app/legacy-app/src/components/pages/News.tsx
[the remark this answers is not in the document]
petebacondarwin: I tried putting invalid search params and it seems to fallback to the first page, no problem...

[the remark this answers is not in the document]
dario-piotrowicz: yeah you're right 😅 

it does happen to work properly but I think it's only because down the line in the fragment we check for the page's validity

you can see that we do end up with `NaN` in the message bus:
Screenshot 2022-11-16 at 13 54 05

<details>
<summary>the 2 `{"page": null}` logs</summary>
the 2 `{"page": null}` logs there are from the news fragment, that's because `JSON.stringify` treats `NaN` as `null`:
Screenshot 2022-11-16 at 13 56 02
</details>

if we remove the check in the fragment things do start to break:
Screenshot 2022-11-16 at 13 58 53

So it is not a big deal (one one ever will notice this anyways I think 😅), if you find the shorter code better I am fine with it, but I sort of feel like the code is not completely correct and we make up for it somewhere else

productivity-suite/app/legacy-app/src/components/pages/News.tsx
dario-piotrowicz: ah, I see 🙂👍

productivity-suite/piercing-library/src/message-bus/get-bus.ts
petebacondarwin: Yeah but I worry that it might break at runtime... if you do `getBus(ref).onCleanup(...)` outside of a fragment, then this will throw a null reference exception.

dario-piotrowicz: ahh.... yeah good point...

productivity-suite/app/legacy-app/src/components/pages/News.tsx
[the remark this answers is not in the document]
petebacondarwin: Even the previous code is not immune to this. For example what about ?page=1.5?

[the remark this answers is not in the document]
dario-piotrowicz: yeah I hadn't considered that 🤦 

I can amend the code to account for that as well (checking that the value is an integer)

But I get the feeling you do prefer the shorter version?
Personally I prefer the longer more robust one (although we are not writing a production application... 😅), but I am definitely not too set on it, I am totally fine with the shorter version (especially since it makes for simpler code in the demo, and we are also already covered as we mentioned)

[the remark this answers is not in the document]
petebacondarwin: OK here is my proposed version:

```ts
function x(v) {
  const n = parseInt(v);
  return n > 0 ? n : 1;
}
```

This gives

```
> x()
1
> x("")
1
> x("foo")
1
> x("-1")
1
> x("-2")
1
> x("2.5")
2
> x("0")
1
```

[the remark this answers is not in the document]
dario-piotrowicz: sounds good 😃👍

productivity-suite/app/legacy-app/src/worker/script.ts
```diff
@@ -50,6 +50,18 @@ const gateway = new PiercingGateway<Env>({
 
       requestMessageBusState["todo-list-selected"] = { name: listName };
     }
+
+    if (!("news-page" in requestMessageBusState)) {
```
petebacondarwin, line 54: I'm not sure we need this if statement, but I don't think it does any harm.

productivity-suite/piercing-library/src/piercing-fragment-host/piercing-fragment-host.ts
```diff
@@ -29,8 +29,9 @@ export class PiercingFragmentHost extends HTMLElement {
 
   disconnectedCallback() {
     if (this.cleanup) {
-      // Only remove handlers if we are actually in cleanup mode.
-      this[messageBusProp].clearAllHandlers();
+      // Only clean up the message bus if we are actually in cleanup mode
+      // and not just piercing this fragment.
+      this[messageBusProp].cleanup();
```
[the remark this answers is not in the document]
petebacondarwin: But having moved that functionality to the fragmenthost, there is actually no need for the renaming now...

productivity-suite/app/legacy-app/src/worker/script.ts
dario-piotrowicz: yeah I agree I don't think it's needed, I included it for consistency with the previous ones, but I think it can be just removed, I don't have a preference on this one... I don't know if having this consistent and drawing a pattern is valuable anyways... feel free to remove it if you want 🙂

productivity-suite/app/fragments/news/src/routes/[...news].tsx
```diff
@@ -3,23 +3,47 @@ import {
...
+  onMount(() => {
+    getBus(ref).listen<{ page: number }>("news-page", ({ page }) => {
+      if (isFinite(page) && page > 0) {
+        setPage(page);
+      }
+    });
+  });
+
+  onMount(() => {
+    getFragmentHost(ref)?.onCleanup(() =>
```
Comment (dario-piotrowicz) [inline productivity-suite/app/fragments/news/src/routes/[...news].tsx:30]: In this case the fragment host should always be present, so it wouldn't be more correct to have:
```suggestion
    getFragmentHost(ref)!.onCleanup(() =>
```
?

Comment (petebacondarwin) [reply productivity-suite/app/fragments/news/src/routes/[...news].tsx:30]: Yes, I guess this is similar to our discussion today, where we expect fragments to always be run inside a context that has a message bus and a host...

productivity-suite/app/legacy-app/src/worker/script.ts
petebacondarwin: Let's leave it in for now.

petebacondarwin (approved):

PR #81, opened 2022-11-16
feat(productivity-suite): slightly improve selected-list delete-btn

dario-piotrowicz:
improve the selected-list delete button so that it doesn't overflow the container nor does it need to have a big padding or font-size

for this fix the selected-list has been converted to a div instead of a button which was apparently making lighthouse considering all its content text which should not be overlapped

by having it as a div we avoid such issue (although this is slightly worse a11y-wise)

___

Before:
Screenshot 2022-11-16 at 10 34 39

After:
Screenshot 2022-11-16 at 10 34 21

(The SEO score is still at 100:
Screenshot 2022-11-16 at 10 35 07)

(the 92 for best-practices happens locally because we get the `rum` error in the console)

### DIFF
--- a/productivity-suite/app/fragments/todo-lists/src/components/SelectedListCard.tsx
+++ b/productivity-suite/app/fragments/todo-lists/src/components/SelectedListCard.tsx
@@ -24,7 +24,7 @@ export const SelectedListCard = component$(
   }) => {
     useStylesScoped$(styles);
 
-    const newTodoInputState = useStore<{
+    const todoListInputState = useStore<{
       editing: boolean;
       value: string | null;
       valid: boolean;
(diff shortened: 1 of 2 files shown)


productivity-suite/app/fragments/todo-lists/src/components/SelectedListCard.tsx
```diff
@@ -24,7 +24,7 @@ export const SelectedListCard = component$(
   }) => {
     useStylesScoped$(styles);
 
-    const newTodoInputState = useStore<{
+    const todoListInputState = useStore<{
```
dario-piotrowicz, line 27: I think this const name wasn't updated after some code change 😓

petebacondarwin (approved):
Cool! OK so this is totally cheating the Lighthouse SEO rule, since we now do actually have the problem that it is more difficult for a phone device to know you have clicked the X button rather than the todo-list item itself.
But I actually think that it is more semantically correct for the item not to be a button.
So let's go with this! Thanks.

PR #82, opened 2022-11-16
Productivity suite - fix body overflow

dario-piotrowicz:
on certain screen sizes we get an ugly overflow, I'm fixing it in this PR
Screenshot 2022-11-16 at 11 05 36

I previously missed it, sorry about that 😓


petebacondarwin (approved):
LGTM - are there any actual devices that are 428px wide?

dario-piotrowicz:
> LGTM - are there any actual devices that are 428px wide?

I am not sure 😅, but this change doesn't break anything, just improves stuff, so either way why not 😄

PR #84, opened 2022-11-17
fix(productivity-suite): fix the todo-list deletion buggy behavior

dario-piotrowicz:
fix the buggy behavior we get when we delete a list in the todo-lists fragment by keeping the root todoLists up to date and slightly tweaking the animation

### DIFF
--- a/productivity-suite/app/fragments/todo-lists/src/components/ListsCarouselAnimations.css
+++ b/productivity-suite/app/fragments/todo-lists/src/components/ListsCarouselAnimations.css
@@ -7,7 +7,7 @@
 }
 
 .todo-list-card {
-    animation-duration: 0.15s;
+    animation-duration: 0.13s;
     animation-timing-function: cubic-bezier(.55,.51,.47,.99);;
 }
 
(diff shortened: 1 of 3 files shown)


petebacondarwin (approved):
I haven't tested this locally but it looks good for now. I am a bit worried that this code seems to be getting quite complicated for what feels like a simple behaviour...

dario-piotrowicz:
> I haven't tested this locally but it looks good for now. I am a bit worried that this code seems to be getting quite complicated for what feels like a simple behaviour...

yeah I totally agree.... yesterday I was trying to simplify things but I was hitting Qwik issues (like it doesn't seem like children get props updates from their parents in certain situations)

on hindsight I regret splitting the component in different sub-components as I feel this just made the code more complex 😓 

(also I implemented the animation using keyframes hoping to keep things simpler... if I had used the web animations api although less known it would have likely produced simpler code 😓)

if you're ok with it let's merge this PR, and if you want I do not mind then spending some extra time (even after the blog post is out) trying to simplify the code here 🙂

PR #86, opened 2022-11-17
fix(productivity-suite): ensure that we only run side-effects once per fragment

petebacondarwin:
Previously, if an outlet was added and then removed quickly, before the fragment had completed its execution, we still added it to the set of previously unmounted fragments. This meant that when the outlet got added a second time, we would manually run the side-effects, even though they would still be automatically run as part of the first-time load.

Now we only add fragments to the previously unmounted set if it completed its execution.

### DIFF
--- a/productivity-suite/piercing-library/src/piercing-fragment-outlet.ts
+++ b/productivity-suite/piercing-library/src/piercing-fragment-outlet.ts
@@ -18,23 +18,25 @@ export function registerPiercingFragmentOutlet() {
 /**
  * Set of ids of fragments that have been previously unmounted.
  *
- * We keep track of this for a single purpose, preventing the
- * extra running of type module scripts for newly fetched fragments.
+ * We keep track of this to know whether to manually run side-effects in modules
+ * that are referenced as part of the fragment.
  *
- * (Meaning that when we fetch a fragment, we check if its id is in the
- * set, if it is then we need to run its type module scripts manually,
- * otherwise the scripts get normally evaluated by the browser).
- *
+ * When we fetch a fragment for the first time any side-effects in module-type scripts
+ * that are referenced will be executed.
+ * Later, if we fetch the fragment again, its side-effects will not run automatically.
+ * To workaround this we ensure that all such modules also expose their side-effects
+ * via a default export function.
+ * We then manually call that default export to re-execute the side-effects.
  */
 const unmountedFragmentIds: Set<string> = new Set();
 
 export class PiercingFragmentOutlet extends HTMLElement {
   // this field can be read from the outside to check if this element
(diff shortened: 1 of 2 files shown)


productivity-suite/piercing-library/src/piercing-fragment-outlet.ts
```diff
@@ -50,21 +52,22 @@ export class PiercingFragmentOutlet extends HTMLElement {
...
+      this.innerHTML == "";
+      this.fragmentHost.pierceInto(this);
     } else {
-      const fragmentStream = await this.fetchFragmentStream();
-      await this.streamFragmentIntoOutlet(fragmentStream);
-      fragmentHost = this.getFragmentHost(true);
+      // We need to fetch and create the fragment host
+      const fragmentStream = await this.fetchFragmentStream(fragmentId);
+      await this.streamFragmentIntoOutlet(fragmentId, fragmentStream);
+      this.fragmentHost = this.getFragmentHost(fragmentId, true);
```
petebacondarwin, line 65: The key point here is that `this.fragmentHost` is only set after the `streamIntoOutlet()` method completed.

productivity-suite/piercing-library/src/piercing-fragment-outlet.ts
```diff
@@ -76,16 +79,14 @@ export class PiercingFragmentOutlet extends HTMLElement {
   }
 
   disconnectedCallback() {
-    unmountedFragmentIds.add(this.fragmentId);
-  }
-
-  private pierceFragmentIntoDOM(fragmentHost: PiercingFragmentHost) {
-    this.innerHTML == "";
-    fragmentHost.pierceInto(this);
+    if (this.fragmentHost) {
```
petebacondarwin, line 82: If there is no `fragmentHost` then we never completed the loading of the fragment.

productivity-suite/piercing-library/src/piercing-fragment-outlet.ts
```diff
@@ -18,23 +18,25 @@ export function registerPiercingFragmentOutlet() {
...
+ * To workaround this we ensure that all such modules also expose their side-effects
+ * via a default export function.
+ * We then manually call that default export to re-execute the side-effects.
  */
 const unmountedFragmentIds: Set<string> = new Set();
 
 export class PiercingFragmentOutlet extends HTMLElement {
   // this field can be read from the outside to check if this element
   // is a PiercingFragmentOutlet (without relying on `instanceof`)
-  // @ts-ignore - the following field is not only being used by the `isPiercingFragmentOutlet` guard
```
dario-piotrowicz, line 34: 😓 👍 (sorry I must have missed this during refactoring)

dario-piotrowicz (approved):
awesome!!! 🤩

PR #87, opened 2022-11-17
fix(productivity-suite): apply cloudflare web analytics only on prod

dario-piotrowicz:
locally we get CORS errors because we are including the cloudflare web-analitics also locally, prevent such behavior by only applying the web analytics if the origin is the prod one

___

This fixes the errors we get locally:
Screenshot 2022-11-17 at 10 59 24

### DIFF
--- a/productivity-suite/app/legacy-app/index.html
+++ b/productivity-suite/app/legacy-app/index.html
@@ -360,17 +360,5 @@
       enableSeamsInput.checked = initialSeamsEnabledValue;
       toggleSeams(initialSeamsEnabledValue);
     </script>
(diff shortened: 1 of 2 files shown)


petebacondarwin (approved):
Thanks @dario-piotrowicz - LGTM.

Is it OK for the script to appear at the top of the head?

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <script defer src="https://static.cloudflareinsights.com/beacon.min.js" data-cf-beacon="{'token': '43a08321febf43fe9103890be2f3bb5f'}"></script>
```

petebacondarwin (approved):

PR #89, opened 2022-11-17
Revert "fix(productivity-suite): apply cloudflare web analytics only on prod

dario-piotrowicz:
Sorry I think this is not working properly 😓 (we may revert it back to the original version or think of some alternative solution)


petebacondarwin:
Hmm, yeah I can see that the script tag appears but then the `rum` requests are not being made. Perhaps having the script in the `head` rather than the `body` makes the difference?

dario-piotrowicz:
> Hmm, yeah I can see that the script tag appears but then the `rum` requests are not being made. Perhaps having the script in the `head` rather than the `body` makes the difference?

ah no, it's not that, it's actually a very silly mistake I made 😓 🙇 

I had to use single quotes for the `data-cf-beacon`: `"{'token': '43a08321febf43fe9103890be2f3bb5f'}"`, I knew that it wasn't valid JSON but hoped that `parse` would still work with it (I thought I checked) but it does not, so I think the whole issue relies there... (and if the script needs to go in the body, I checked and, Vite's `HtmlTagDescriptor` does allow us to specify if the tag should be injected in the body, so that's not really an issue)

the fact is that Vite doesn't seem to handle double quotes in attributes correctly (since it sets attributes with double quotes), so when I try to use double quotes this is the result Vite produces:
Screenshot 2022-11-18 at 09 41 41
which doesn't seem to work either....

I'll try to see if there is a way to use double quotes there

(another solution if we want to go there is to change the plugin to actually manually update the html, it's much less nice but it should do the trick)

dario-piotrowicz:
Ok, I've fixed it by using `&quot;`, in the Vite plugin it looks rather awkward 😓, but it does produce the correct html

PR #95, opened 2022-11-28
Productivity suite running legacy app

petebacondarwin:


productivity-suite/piercing-library/src/message-bus/get-bus.ts
```diff
@@ -1,13 +1,17 @@
 import { getFragmentHost } from "../piercing-fragment-host/get-fragment-host";
-import type { MessageBus } from "./message-bus";
+import { GenericMessageBus, type MessageBus } from "./message-bus";
 import { messageBusProp } from "./message-bus-prop";
 
 /**
- * Gets a message bus based on the provided element.
+ * Get a `MessageBus` based on the provided element.
```
dario-piotrowicz, line 6: I think that js/ts docs (unlike commit messages) should be in present simple rather than imperative for example:

Screenshot 2022-11-28 at 10 51 33

Screenshot 2022-11-28 at 10 52 12

productivity-suite/piercing-library/src/message-bus/get-bus.ts
```diff
@@ -21,9 +25,16 @@ export function getBus(element?: Element): MessageBus {
   if (hasMessageBus(root)) {
     return root[messageBusProp];
   }
-  throw new Error("No global message bus defined!");
+  console.warn("No global message bus defined - creating an empty one.");
```
dario-piotrowicz, line 28: I am not completely sure about the warn, if you run a fragment in isolation this is what you actually want right? so always displaying the warning anyways feels a bit off to me 😕

productivity-suite/piercing-library/src/message-bus/get-bus.ts
```diff
@@ -21,9 +25,16 @@ export function getBus(element?: Element): MessageBus {
...
+
+  return ((globalThis as unknown as { [messageBusProp]?: GenericMessageBus })[
+    messageBusProp
+  ] = new GenericMessageBus());
 }
 
+/**
+ * Is the given `obj` a `MessageBus` container?
+ */
```
dario-piotrowicz, line 37: I am not too sure if this comment is adding much 😕

dario-piotrowicz (approved):
Nice! 😃

productivity-suite/piercing-library/src/message-bus/get-bus.ts
petebacondarwin: Yeah - I guess so. I just find it ugly. But I should be consistent with the rest of the world 😄

productivity-suite/piercing-library/src/message-bus/get-bus.ts
petebacondarwin: If you are running the app in isolation then you are going to be getting placeholders instead of real fragments, which seems to me not a production scenario.

productivity-suite/piercing-library/src/message-bus/get-bus.ts
petebacondarwin: I added another line!

productivity-suite/piercing-library/src/message-bus/get-bus.ts
dario-piotrowicz: 😅👍 

anyways I guess it could be a bit ugly when looking at it in the code, but I find that it does make sense when you see it as IDE suggestions since they are explaining stuff to the developer, imperative would look ugly there, wouldn't it? 🤔

productivity-suite/piercing-library/src/message-bus/get-bus.ts
dario-piotrowicz: yes but you'd still get the warning during development right?

petebacondarwin: That's right! And I think that is a good thing. IMO.

dario-piotrowicz: ah ok 🙂👍

PR #96, opened 2022-12-02
fix(productivity-suite): add missing text-align center for todo-list cards

dario-piotrowicz:
before:
Screenshot 2022-12-02 at 17 59 10

after:
Screenshot 2022-12-02 at 17 58 36


petebacondarwin (approved):
LGTM

PR #98, opened 2022-12-12
Add a reframed demo

jculvey:


dario-piotrowicz (commented):
I agree with @petebacondarwin that the reframed lib and the piercing one should be separate

also the piercingGateway is exactly the same so I think this would have the issue I mentioned regarding the assets and js getting loaded in the main document?

(PS: I know we are going to iterate over this so I am ok if we want to merge it in as is, by the way, I've still got to review the reframed-client file 🙂)

PR #101, opened 2022-12-15
Repierced demo

jculvey:
This is a rework of https://github.com/cloudflare/workers-web-experiments/pull/98 that retains piercing.


productivity-suite/piercing-library/src/reframed-client.js
```diff
@@ -0,0 +1,241 @@
...
+// all frames contribute to the joint history: https://www.w3.org/TR/2011/WD-html5-20110525/history.html#joint-session-history
+// this means that we need to be careful not to add duplicate entries to the
+// history stack via pushState within the iframe as that would double the
+// number of history entries that back/forward button would have to work through
+//
+// therefore we do the following:
+// - intercept all history.pushState history.replaceState calls and replay
+//   them in the main window
+// - update the window.location within the iframe via history.replaceState
+// - intercept window.addEventListener('popstate', ...) registration and forward it onto the main window
```
dario-piotrowicz, line 163: I don't think we do anything like this in the demo? could we strip the history logic out so to make the code simpler? (and mention this possible implementation in an md file)

productivity-suite/piercing-library/src/reframed-client.js
```diff
@@ -0,0 +1,241 @@
...
+  originalHistoryFns.set(prop, window.history[prop]);
+  Object.defineProperty(window.history, prop, {
+    get: () => {
+      return function reframedHistoryGetter() {
+        console.log(
+          prop,
+          "history length",
+          reframedWindow.history.length,
+          window.history.length
+        );
```
dario-piotrowicz, line 175: is this console log here on purpose?

productivity-suite/piercing-library/src/reframed-client.js
```diff
@@ -0,0 +1,241 @@
...
+
+        switch (prop) {
+          case "pushState": {
+            Reflect.apply(
```
dario-piotrowicz, line 179: Nit: is `Reflect` necessary? above we're not using it and just doing `.apply`, could we keep that for consistency?

productivity-suite/piercing-library/src/reframed-client.js
```diff
@@ -0,0 +1,241 @@
...
+    reframedWindow.history.state,
+    null,
+    reframedWindow.location.href,
+  ]);
+});
+
+["length", "scrollRestoration", "state"].forEach((prop) => {
+  Object.defineProperty(window.history, prop, {
+    get: () => {
+      return Reflect.get(reframedWindow.history, prop);
```
dario-piotrowicz, line 221: is the `Reflect.get` necessary here? wouldn't `reframedWindow.history[prop]` work just as well?

productivity-suite/piercing-library/src/stream-utilities.ts
```diff
@@ -21,6 +21,40 @@ export function wrapStreamInText(
   return readable;
 }
 
+export function transformStream(
```
dario-piotrowicz, line 24: I think it's sort of explanatory, but since this is exported (so I guess it's public api) could you add a tsdoc for this function like the one in the function above? 🙏

dario-piotrowicz (approved):

productivity-suite/piercing-library/src/piercing-gateway.ts
```diff
@@ -330,19 +353,41 @@ export class PiercingGateway<Env> {
...
+    const framework = fragmentConfig.framework;
+
+    const prePiercingStyles = prePiercing
+      ? `<style>${fragmentConfig.prePiercingStyles}</style>`
+      : ``;
+
+    let template = `
+      <piercing-fragment-host fragment-id=${fragmentConfig.fragmentId}>
+        ${prePiercingStyles}
+        --FRAGMENT_CONTENT--
```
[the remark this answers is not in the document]
jculvey: Agreed it's a little weird, but it's just a delimiter we use to split the fragment on, so we can make it anything really. It's place in the world is similar to something you'd see in a server side templating language, like `<% fragment_content %>`

I chose this because I kinda wanted it to stick out as specifically not HTML. I was worried that folks might confuse it for actual markup that would be on the page. `FRAGMENT_CONTENT` could be a constant, and #FRAGMENT_CONTENT might conceivably be a token found in `prePiercingStyles`, which would cause issues. By using something that is invalid HTML, we run less of a risk that the `.split` call on template fails.

productivity-suite/piercing-library/src/reframed-client.js
```diff
@@ -0,0 +1,241 @@
+const reframedContainer = window.parent.document.querySelector(
```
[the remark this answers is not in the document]
jculvey: It's a little more convenient as js. 

This isn't like other .ts files -- it's a script sort of like the qwikloader that we're directly inlining into the fragment, not `require`-ing in some other part of the library. So even if we typed it, the safety benefit would only be from within the file, not on callsites that consume this code (since there aren't any). I'm guessing it would also mean that requiring it with `?raw` would return typescript, which would need to get transpiled before it could be injected into the page. And the tooling around that seems a little annoying.

productivity-suite/piercing-library/src/reframed-client.js
jculvey: Not sure, @IgorMinar does `Reflect` give us anything special here?

productivity-suite/piercing-library/src/reframed-client.js
jculvey: @IgorMinar thoughts on this?

productivity-suite/piercing-library/src/piercing-gateway.ts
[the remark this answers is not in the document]
dario-piotrowicz: I see, yeah that makes sense, thanks for the clarification 🙂

productivity-suite/piercing-library/src/reframed-client.js
```diff
@@ -0,0 +1,234 @@
...
+  const originalDocumentFn = document[listenerProperty];
+  document[listenerProperty] = function reframedListenerFn(eventName) {
+    if (eventName === "DOMContentLoaded") {
+      return originalDocumentFn.apply(document, arguments);
+    }
+    return reframedContainer[listenerProperty].apply(
+      reframedContainer,
+      arguments
+    );
+  };
```
dario-piotrowicz, line 112:
```suggestion
  document[listenerProperty] = function reframedListenerFn(eventName, ...args) {
    if (eventName === "DOMContentLoaded") {
      return originalDocumentFn.apply(document, [eventName, ...args]);
    }
    return reframedContainer[listenerProperty].apply(
      reframedContainer,
      [eventName, ...args]
    );
  };
```

or maybe:
```suggestion
  document[listenerProperty] = function reframedListenerFn(...args) {
    const eventName = args[0];
    if (eventName === "DOMContentLoaded") {
      return originalDocumentFn.apply(document, args);
    }
    return reframedContainer[listenerProperty].apply(
      reframedContainer,
      args
    );
  };
```

productivity-suite/piercing-library/src/reframed-client.js
```diff
@@ -0,0 +1,234 @@
...
+// - intercept all history.pushState history.replaceState calls and replay
+//   them in the main window
+// - update the window.location within the iframe via history.replaceState
+// - intercept window.addEventListener('popstate', ...) registration and forward it onto the main window
+const originalHistoryFns = new Map();
+["back", "forward", "go", "pushState", "replaceState"].forEach((prop) => {
+  originalHistoryFns.set(prop, window.history[prop]);
+  Object.defineProperty(window.history, prop, {
+    get: () => {
+      return function reframedHistoryGetter() {
```
dario-piotrowicz, line 162:
```suggestion
      return function reframedHistoryGetter(...args) {
```

productivity-suite/piercing-library/src/reframed-client.js
```diff
@@ -0,0 +1,234 @@
...
+            );
+            reframedWindow.history.replaceState(...arguments);
+            break;
+          }
+          default: {
+            Reflect.apply(
+              reframedWindow.history[prop],
+              reframedWindow.history,
+              arguments
+            );
```
dario-piotrowicz, line 194:
```suggestion
              args
            );
            reframedWindow.history.pushState(...args);
            break;
          }
          case "replaceState": {
            Reflect.apply(
              originalHistoryFns.get("replaceState"),
              window.history,
              args
            );
            reframedWindow.history.replaceState(...args);
            break;
          }
          default: {
            Reflect.apply(
              reframedWindow.history[prop],
              reframedWindow.history,
              args
            );
```

productivity-suite/piercing-library/src/reframed-client.js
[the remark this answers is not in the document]
dario-piotrowicz: ah yeah I see, I did what you mentioned here: https://github.com/cloudflare/workers-web-experiments/blob/df50b60cfff7bc299cf70ecfe8f7826ec9313b84/productivity-suite/piercing-library/message-bus.vite.config.ts

having the js transpiled is quite nice as we can also minify it, but it is a bit of a headache/overhead 😓, so I'm ok with whatever you think is best 🙂

productivity-suite/piercing-library/src/reframed-client.js
petebacondarwin: Not sure what benefit you get from Reflect here...

productivity-suite/piercing-library/src/reframed-client.js
petebacondarwin: I don't think so...

productivity-suite/piercing-library/src/reframed-client.js
IgorMinar: Ok to strip for now

productivity-suite/piercing-library/src/reframed-client.js
IgorMinar: Debugging only. Can be removed

productivity-suite/piercing-library/src/reframed-client.js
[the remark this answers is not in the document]
IgorMinar: JS is fine for now. Alternatively we can switch to TS and minify.

productivity-suite/piercing-library/src/reframed-client.js
IgorMinar: I tried to use Reflect everywhere where I was accessing the foreign context. This was done for improving code readability. Metaprogramming and cross context code can be hard to read that's why I used Reflect api as an aid.

IgorMinar: I don't recall any part of this code actually requiring Reflect apis. But if we stick with it we should use it consistently.

IgorMinar: One more reason for Reflect was that it's also consistent with the use of Object.defineProperty pattern in this file.

PR #102, opened 2023-01-26
Fix memory leaks

jculvey:
Addresses the memory leak issue we found in the reframed demo

### DIFF
--- a/productivity-suite/app/fragments/todos/src/components/Header.tsx
+++ b/productivity-suite/app/fragments/todos/src/components/Header.tsx
@@ -26,6 +26,8 @@ export function Header({
         }`}
         placeholder="What needs to be done?"
         autoFocus
+        autoComplete="off"
+        spellCheck={false}
         value={newTodoDetails.text}
         onInput={(event: ChangeEvent<HTMLInputElement>) => {
           const text = event.target.value;
--- a/productivity-suite/piercing-library/src/message-bus/message-bus.ts
+++ b/productivity-suite/piercing-library/src/message-bus/message-bus.ts
@@ -60,6 +60,7 @@ export class GenericMessageBus implements MessageBus {
   }
 
   dispatch(eventName: string, value: JSONValue) {
+    value = JSON.parse(JSON.stringify(value));
     this._state[eventName] = value;
     const callbacksForEvent = this._callbacksMap.get(eventName) ?? [];
     setTimeout(
--- a/productivity-suite/piercing-library/src/piercing-fragment-outlet.ts
+++ b/productivity-suite/piercing-library/src/piercing-fragment-outlet.ts
@@ -80,7 +80,28 @@ export class PiercingFragmentOutlet extends HTMLElement {
 
   disconnectedCallback() {
     if (this.fragmentHost) {
-      unmountedFragmentIds.add(this.fragmentHost.fragmentId);
+      const fragmentId = this.fragmentHost.fragmentId;
+      unmountedFragmentIds.add(fragmentId);
+
+      const fragmentFunctionConstructor =
+        fragmentRegistrationMap.get(fragmentId)!;
+      if (fragmentFunctionConstructor) {
+        fragmentRegistrationMap.delete(fragmentId);
+
+        const listenerRegistrations = fragmentListenerMap.get(
+          fragmentFunctionConstructor
+        )!;
+        fragmentListenerMap.delete(fragmentFunctionConstructor);
+
+        for (const listenerRegistration of listenerRegistrations) {
+          listenerRegistration.target.removeEventListener(
+            listenerRegistration.name,
+            listenerRegistration.listener,
+            listenerRegistration.options
+          );
+        }
+      }
+
       this.fragmentHost = null;
     }
   }
--- a/productivity-suite/piercing-library/src/piercing-gateway.ts
+++ b/productivity-suite/piercing-library/src/piercing-gateway.ts
@@ -374,12 +374,13 @@ export class PiercingGateway<Env> {
     if (this.config.isolateFragments?.(env)) {
       // Quotes must be escaped from srcdoc contents
       template = `
+        ${prePiercingStyles}
         <piercing-fragment-host fragment-id=${fragmentConfig.fragmentId}>
         </piercing-fragment-host>
-        ${prePiercingStyles}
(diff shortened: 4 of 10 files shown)


productivity-suite/app/fragments/todos/src/components/Header.tsx
```diff
@@ -26,6 +26,8 @@ export function Header({
         }`}
         placeholder="What needs to be done?"
         autoFocus
+        autoComplete="off"
```
IgorMinar, line 29: I believe that this attribute makes no difference. Can you please confirm and remove?

productivity-suite/app/fragments/todos/src/components/Header.tsx
```diff
@@ -26,6 +26,8 @@ export function Header({
...
+        spellCheck={false}
```
IgorMinar, line 30: Can you please add a comment with a reference to the chrome big we filed? Thanks

productivity-suite/piercing-library/src/message-bus/message-bus.ts
```diff
@@ -60,6 +60,7 @@ export class GenericMessageBus implements MessageBus {
   }
 
   dispatch(eventName: string, value: JSONValue) {
+    value = JSON.parse(JSON.stringify(value));
```
IgorMinar, line 63: Can you please add a comment to explain what's going on here?

Also could we use https://developer.mozilla.org/en-US/docs/Web/API/structuredClone instead? (structuredClone feels less hacky)

productivity-suite/piercing-library/src/piercing-fragment-outlet.ts
```diff
@@ -80,7 +80,28 @@ export class PiercingFragmentOutlet extends HTMLElement {
 
   disconnectedCallback() {
     if (this.fragmentHost) {
-      unmountedFragmentIds.add(this.fragmentHost.fragmentId);
+      const fragmentId = this.fragmentHost.fragmentId;
```
IgorMinar, line 83: Can you add a comment before this block explaining what we are doing here? Thanks

productivity-suite/piercing-library/src/piercing-gateway.ts
```diff
@@ -372,12 +374,13 @@ export class PiercingGateway<Env> {
     if (this.config.isolateFragments?.(env)) {
       // Quotes must be escaped from srcdoc contents
       template = `
+        ${prePiercingStyles}
         <piercing-fragment-host fragment-id=${fragmentConfig.fragmentId}>
         </piercing-fragment-host>
-        ${prePiercingStyles}
```
IgorMinar, line 377: Ahh... I see. This explains the FOUC

productivity-suite/piercing-library/src/piercing-gateway.ts
```diff
@@ -463,6 +466,35 @@ function getEscapedReframedClientCode(fragmentId: string) {
   </script>`;
 }
 
+function getEmbeddedStyleScript(fragmentId: string) {
```
IgorMinar, line 469: Can you please document the purpose of this fn. thank you

productivity-suite/piercing-library/src/reframed-host.js
```diff
@@ -0,0 +1,54 @@
+// Maps that allow us to keep track of registered fragments and their listeners.
+// We use FunctionConstructor as a way to identify the origin of a registered listener and map it back to its fragment.
+const fragmentRegistrationMap = new Map();
+const fragmentListenerMap = new Map();
+
+// monkey patch global addEventListner on window and document so that we can automatically unregister
+// any listeners created from within a registered fragment.
+const addEventListenerTargets = [window, document];
+// const addEventListenerTargets = [Node.prototype];
```
IgorMinar, line 9: Remove?

productivity-suite/piercing-library/src/reframed-host.js
```diff
@@ -0,0 +1,54 @@
...
+        listenerFnConstructor = listener.constructor;
+      }
+
+      const listenerFns = fragmentListenerMap.get(listenerFnConstructor);
+
+      // if this listener comes from a registered fragment, then keep track of it so that we can unregister it
+      if (listenerFns) {
+        listenerFns.push({
+          target: addEventListenerTarget,
+          // target: this,
```
IgorMinar, line 28: Remove

productivity-suite/piercing-library/src/reframed-host.js
```diff
@@ -0,0 +1,54 @@
...
+          name,
+          listener,
+          options,
+        });
+      }
+
+      return originalDocumentAddEventListener.call(
+        addEventListenerTarget,
+        // this,
```
IgorMinar, line 37: Remove

IgorMinar (approved):
Lgtm once the review feedback is addressed! Thank you James.

productivity-suite/piercing-library/src/piercing-gateway.ts
```diff
@@ -380,6 +381,7 @@ export class PiercingGateway<Env> {
         <iframe id="iframe_${fragmentId}" style="display: none" srcdoc="
           <body>
             --FRAGMENT_CONTENT--
+            <div class="fragment-end"></div>
```
IgorMinar, line 384: same here

productivity-suite/piercing-library/src/piercing-gateway.ts
```diff
@@ -368,6 +368,7 @@ export class PiercingGateway<Env> {
       <piercing-fragment-host fragment-id=${fragmentConfig.fragmentId}>
         ${prePiercingStyles}
         --FRAGMENT_CONTENT--
+        <div class="fragment-end"></div>
```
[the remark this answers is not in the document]
jculvey: I can use `body.show-seams piercing-fragment-host:has(*)` instead. Dario mentioned a corner case where the fragment host contains non-visible elements (which would still cause the seams to display). I can't repro that with the productivity suite, but it can definitely happen.

I can't really think of an alternative to the above selector that wouldn't involve embedding a sentinel value element that we then query for with a more specific selector. Open to ideas.

Personally, I vote to go with `body.show-seams piercing-fragment-host:has(*)` for now since it makes the demo code simpler unless we can find a use case within the productivity suite which breaks things.

